Issue #22, opened 2016-03-05
More convenient Iota wrapper


wilzbach:
Looking throw the unittests I see this pattern coming up > 80% of all tests.

```
auto b = 4.iota.sliced(2, 2);
auto tensor = 60.iota.array.sliced(3, 4, 5);
auto a = 240000.iota.sliced(10, 20, 30, 40);
```

I do like this pattern, but seeing 1) how often this occurs and 2) that you need to know the size of iota no that `downSize` is disabled by default it seems to me that having a convenience wrapper for this, like `Iota`.

```
Iota.sliced(10, 20, 30, 40);
```

It looks nicer and avoids the problem that you actually mistyped the number of elements - which can only be evaluated in runtime - e.g. `24000.iota.sliced(10, 20, 30, 40);` will result in an hard-to-see runtime error.

I haven't thought much about how we could combine this `start, stepsize` from usual iota, but how about:

```
Iota!10(10,20,30,40) // starts at ten
Iota!(10, 2)(10,20,30,40) // starts at ten, with steps of 2
Iota!(10, 0.5)(10,20,30,40) // starts at ten, with steps of 0.5
```

Would be happy to send a PR for this ;-)

9il:
No allocations, but the same logic (compute total length from dimensions). So this is for the future allocation module too

wilzbach:
@9il Let me know your progress on the future allocation module and how I can help you moving this amazing library forward ;-)

9il:
@greenify If you would have a time to improve documentation or write an article for Numpy users (with compression like in #24) this would be really helpful!

wilzbach:
Sure I will do that, but I was more asking for coding tasks ;-)

wilzbach:
@9il I copied all function headers from NumPy into the [wiki](https://github.com/DlangScience/mir/wiki/NumPy), so we can start by tagging
- what is implemented and how it can be achieved with `mir`
- what is in process (just rough areas, e.g. I think you already have a stats module)

wilzbach:
@9il as `squeeze` didn't turn out that well, what other functions/feature do you want to bring to mir in the future?

Imho [Array creation](https://github.com/DlangScience/mir/wiki/NumPy:-Array-creation), [Array manipulation](https://github.com/DlangScience/mir/wiki/NumPy:-Array-manipulation-routines) 
[...]

PR #27, opened 2016-03-05
notice about the NumPy - mir wiki entries

wilzbach:
as mentioned in issue #22

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,3 +13,7 @@ Numeric library and mirror for upcoming numeric packages for the Dlang standard 
 
 ## Packages
  - `ndslice` [Multidimensional Random Access Ranges and Arrays](http://dlang.org/phobos-prerelease/std_experimental_ndslice.html)
+
+## Comparison with NumPy
+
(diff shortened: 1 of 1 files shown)


9il:
Thank you for the wiki! First we should feel already existing features.
For example `rollaxis` analog is `x.pack!2.evertPack.unpack`. `a = b` works too (if I am not wrong). Plus https://github.com/DlangScience/mir/wiki/NumPy:-Indexing can be easily implemented by user (see examples from https://dlang.org/phobos/std_experimental_ndslice_selection.html).

wilzbach:
@9il - agreed that we shouldn't promote this as long as it is still in work ;-)
But let's keep this PR open, so that we don't forget it and can track our progress?

9il:
> But let's keep this PR open, so that we don't forget it and can track our progress?

of course

wilzbach:
I changed my mind - let's put this as a task (see #77) and not as a PR ;-)

PR #29, opened 2016-03-14
mir.combinatorics: added binomial, cartesianPower, combinations, combinationsRepeat, permuations

wilzbach:
As discussed at https://github.com/D-Programming-Language/phobos/pull/4026#issuecomment-196493903 here comes `bionomial` - the first function for `mir`'s upcoming statistics package ;-)

I didn't know where to put it exactly - so I went ahead and created `stat.d` - feel free to suggest a better name for the module

### DIFF
--- a/source/mir/combinatorics/package.d
+++ b/source/mir/combinatorics/package.d
@@ -8,7 +8,6 @@ License: $(LINK2 http://boost.org/LICENSE_1_0.txt, Boost License 1.0).
 module mir.combinatorics;
 
 import std.range.primitives: isInputRange, hasLength;
-import std.traits: isFloatingPoint;
 
 /**
 Checks whether we can do basic arithmetic operations, comparisons, modulo and
@@ -65,69 +64,58 @@ Returns:
 */
 
 R binomial(R = ulong, T)(T n, T k)
-    if ((isArithmetic!(R, T) || isFloatingPoint!R) &&
+    if (isArithmetic!(R, T) &&
         ((is(typeof(T.min < 0)) && is(typeof(T.init & 1))) || !is(typeof(T.min < 0))) )
 {
-    import std.traits: isFloatingPoint;
-    static if (!isFloatingPoint!R)
-    {
-        R result = 1;
+    R result = 1;
 
-        enum hasMinProperty = is(typeof(T.min < 0));
-        // only add negative support if possible
-        static if ((hasMinProperty && T.min < 0) || !hasMinProperty)
+    enum hasMinProperty = is(typeof(T.min < 0));
+    // only add negative support if possible
+    static if ((hasMinProperty && T.min < 0) || !hasMinProperty)
+    {
+        if (n < 0)
         {
-            if (n < 0)
+            if (k >= 0)
             {
-                if (k >= 0)
-                {
-                    return (k & 1 ? -1 : 1) * binomial!(R, T)(-n + k-1, k);
-                }
-                else if (k <= n)
-                {
-                    return ((n-k) & 1 ? -1 : 1) * binomial!(R, T)(-k-1, n-k);
-                }
+                return (k & 1 ? -1 : 1) * binomial!(R, T)(-n + k-1, k);
             }
-            if (k < 0)
+            else if (k <= n)
             {
-                result = 0;
-                return result;
+                return ((n-k) & 1 ? -1 : 1) * binomial!(R, T)(-k-1, n-k);
             }
         }
-
-        if (k > n)
+        if (k < 0)
         {
             result = 0;
             return result;
         }
-        if (k > n - k)
-        {
-            k = n - k;
-        }
-        // make a copy of n (could be a custom type)
-        for (T i = 1, m = n; i <= k; i++, m--)
-        {
-            // check whether an overflow can happen
-            // hasMember!(Result, "max") doesn't work with dmd2.068 and ldc 0.17
-            static if (is(typeof(0 > R.max)))
-            {
-                if (result / i > R.max / m) return 0;
-                result = result / i * m + result % i * m / i;
-            }
-            else
-            {
-                result = result * m / i;
-            }
-        }
+    }
+
+    if (k > n)
+    {
+        result = 0;
         return result;
     }
-    else
+    if (k > n - k)
     {
-        import std.math: exp;
-        import std.mathspecial: logGamma;
-        real x = exp(logGamma(n+1)-logGamma(k+1)-logGamma(n-k+1));
-        return x;
+        k = n - k;
     }
+    // make a copy of n (could be a custom type)
+    for (T i = 1, m = n; i <= k; i++, m--)
+    {
+        // check whether an overflow can happen
+        // hasMember!(Result, "max") doesn't work with dmd2.068 and ldc 0.17
+        static if (is(typeof(0 > R.max)))
+        {
+            if (result / i > R.max / m) return 0;
+            result = result / i * m + result % i * m / i;
+        }
+        else
+        {
+            result = result * m / i;
+        }
+    }
+    return result;
 }
 
 ///
@@ -177,22 +165,9 @@ unittest
     assert(binomial!BigInt(-5, 3) == -35);
     assert(binomial!BigInt(5, -3) == 0);
     assert(binomial!BigInt(-5, -7) == 15);
-
-    // floating point
-    import std.math: approxEqual;
-    assert(binomial!real(5.8, 3).approxEqual(17.632, 0.001));
-    assert(binomial!real(20.8, 3.2).approxEqual(1783.34, 0.01));
-    assert(binomial!real(-5.3, 4).approxEqual(84.2958375, 0.0001));
-    assert(binomial!real(-5.6, -3.4).approxEqual(0.00998410, 0.00001));
-    assert(binomial!real(-5.7, -7.6).approxEqual(20.1795, 0.0001));
-
-    // big floating point
-    assert(binomial!real(1.2e10, 2.3).approxEqual(5.66794e22, 0.0001));
-    // TODO: rounding errors seem to sum up here
-    //assert(binomial!real(2.245e15, 5.7).approxEqual(7.68342e84, 0.1));
 }
 
-private struct IndexedRoR(Collection, Range)
+struct IndexedRoR(Collection, Range)
 if (isInputRange!Range)
 {
     import std.range : indexed, isForwardRange;
@@ -206,16 +181,19 @@ if (isInputRange!Range)
         this.r = range;
     }
 
+    ///
     @property auto ref front()
     {
         return r.indexed(c.front);
     }
 
+    ///
     void popFront()
     {
         c.popFront;
     }
 
+    ///
     @property bool empty()
     {
         return c.empty;
@@ -223,6 +201,7 @@ if (isInputRange!Range)
 
     static if(hasLength!Collection)
     {
+        ///
         @property size_t length()
         {
             return c.length;
@@ -231,6 +210,7 @@ if (isInputRange!Range)
 
     static if(isForwardRange!Collection)
     {
+        ///
         @property typeof(this) save()
         {
             return IndexedRoR!(Collection, Range)(c.save, r);
@@ -241,6 +221,13 @@ if (isInputRange!Range)
 /**
 Creates a projection of `collection` Range (e.g. `Combinations`)
 onto a custom `range`.
+
+Params:
+    collenction = Range to be projected from
+    range = Range to be projected to
+
+Returns:
+    Range with a projection to range for every element of collection
 */
 IndexedRoR!(Collection, Range) indexedRoR(Collection, Range)(Collection c, Range r)
 if (isInputRange!Range)
@@ -263,7 +250,7 @@ Params:
     alloc = custom Allocator
 
 Returns:
-    Forward range which yields the permutations
+    Forward range, which yields the permutations
 */
 Permutations permutations(size_t n) @safe pure nothrow
 {
@@ -321,7 +308,15 @@ Permutations makePermutations(Allocator)(auto ref Allocator alloc, size_t n)
     return Permutations(state, indices);
 }
 
-/// Dispose a Permutations object
+/*
+Disposes a Permutations object. It destroys and then deallocates the
+Permutations object pointed to by a pointer.
+It is assumed the respective entities had been allocated with the same allocator.
+
+Params:
+    alloc = Custom allocator
+    perm = Permutations object
+*/
 void dispose(Allocator)(auto ref Allocator alloc, auto ref Permutations perm)
 {
     import std.experimental.allocator: dispose;
@@ -344,7 +339,7 @@ static if (__VERSION__ > 2069) @nogc unittest
 }
 
 /// ditto
-private struct Permutations
+struct Permutations
 {
     private uint[] indices, state;
     private bool _empty;
@@ -367,11 +362,13 @@ private struct Permutations
         _empty = indices.length == 0;
     }
 
+    ///
     @property uint[] front() @safe pure nothrow @nogc
     {
         return indices;
     }
 
+    ///
     void popFront() @safe pure nothrow @nogc
     {
         import std.algorithm.mutation : swapAt;
@@ -398,11 +395,13 @@ private struct Permutations
         }
     }
 
+    ///
     @property bool empty() @safe pure nothrow @nogc
     {
         return _empty;
     }
 
+    ///
     @property Permutations save() @safe pure nothrow
     {
         typeof(this) c = this;
@@ -429,7 +428,7 @@ Params:
     alloc = custom Allocator
 
 Returns:
-    Forward range which yields the product items
+    Forward range, which yields the product items
 */
 CartesianPower cartesianPower(size_t n, size_t repeat = 1) @safe pure nothrow
 in
@@ -465,14 +464,23 @@ body
     return CartesianPower(n, alloc.makeArray!uint(repeat));
 }
 
-/// Dispose a CartesianPower object
+/*
+Disposes a CartesianPower object. It destroys and then deallocates the
+CartesianPower object pointed to by a pointer.
+It is assumed the respective entities had been allocated with the same allocator.
+
+Params:
+    alloc = Custom allocator
+    perm = CartesianPower object
+*/
+
 void dispose(Allocator)(auto ref Allocator alloc, auto ref CartesianPower cartesianPower)
 {
     import std.experimental.allocator: dispose;
     dispose(alloc, cartesianPower._state);
 }
 
-private struct CartesianPower
+struct CartesianPower
 {
 
 private:
@@ -493,18 +501,13 @@ public:
         _max_states = pow(n, state.length);
     }
 
-    @property CartesianPower save() @safe pure nothrow
-    {
-        typeof(this) c = this;
-        c._state = _state.dup;
-        return c;
-    }
-
+    ///
     @property uint[] front() @safe pure nothrow @nogc
     {
         return _state;
     }
 
+    ///
     void popFront() @safe pure nothrow @nogc
     {
         assert(!empty);
@@ -529,14 +532,24 @@ public:
         }
     }
 
+    ///
     @property size_t length() @safe pure nothrow @nogc
     {
         return _max_states - _pos;
     }
 
+    ///
     @property bool empty() @safe pure nothrow @nogc
     {
         return _pos == _max_states;
+    }
+
+    ///
+    @property CartesianPower save() @safe pure nothrow
+    {
+        typeof(this) c = this;
+        c._state = _state.dup;
+        return c;
     }
 }
 
@@ -617,7 +630,7 @@ Params:
     alloc = custom Allocator
 
 Returns:
-    Forward range which yields the k-combinations items
+    Forward range, which yields the k-combinations items
 */
 Combinations combinations(size_t n, size_t k = 1) @safe pure nothrow
 in
@@ -653,14 +666,23 @@ body
     return Combinations(cast(uint) n, alloc.makeArray!uint(cast(uint) repeat));
 }
 
-/// Dispose a Combinations object
+/*
+Disposes a Combinations object. It destroys and then deallocates the
+Combinations object pointed to by a pointer.
+It is assumed the respective entities had been allocated with the same allocator.
+
+Params:
+    alloc = Custom allocator
+    perm = Combinations object
+*/
+
 void dispose(Allocator)(auto ref Allocator alloc, auto ref Combinations combs)
 {
     import std.experimental.allocator: dispose;
     dispose(alloc, combs.state);
 }
 
-private struct Combinations
+struct Combinations
 {
 
 private:
@@ -694,11 +716,13 @@ public:
         }
     }
 
+    ///
     @property uint[] front() @safe pure nothrow @nogc
     {
         return state;
     }
 
+    ///
     void popFront() @safe pure nothrow @nogc
     {
         assert(!empty);
@@ -739,16 +763,19 @@ public:
         }
     }
 
+    ///
     @property size_t length() @safe pure nothrow @nogc
     {
         return max_states - pos;
     }
 
+    ///
     @property bool empty() @safe pure nothrow @nogc
     {
         return pos == max_states;
     }
 
+    ///
     @property Combinations save() @safe pure nothrow
     {
         typeof(this) c = this;
@@ -883,7 +910,7 @@ Params:
     alloc = custom Allocator
 
 Returns:
-    Forward range which yields the k-multicombinations items
+    Forward range, which yields the k-multicombinations items
 */
 CombinationsRepeat combinationsRepeat(size_t n, size_t k = 1) @safe pure nothrow
 in
@@ -919,14 +946,23 @@ body
     return CombinationsRepeat(n, alloc.makeArray!uint(repeat));
 }
 
-/// Dispose a CombinationsRepeat object
+/*
+Disposes a CombinationsRepeat object. It destroys and then deallocates the
+CombinationsRepeat object pointed to by a pointer.
+It is assumed the respective entities had been allocated with the same allocator.
+
+Params:
+    alloc = Custom allocator
+    perm = CombinationsRepeat object
+*/
+
 void dispose(Allocator)(auto ref Allocator alloc, auto ref CombinationsRepeat combs)
 {
     import std.experimental.allocator: dispose;
     dispose(alloc, combs.state);
 }
 
-private struct CombinationsRepeat
+struct CombinationsRepeat
 {
 
 private:
@@ -949,11 +985,13 @@ public:
         }
     }
 
+    ///
     @property uint[] front() @safe pure nothrow @nogc
     {
         return state;
     }
 
+    ///
     void popFront() @safe pure nothrow @nogc
     {
         assert(!empty);
@@ -981,16 +1019,19 @@ public:
         }
     }
 
+    ///
     @property size_t length() @safe pure nothrow @nogc
     {
         return max_states - pos;
     }
 
+    ///
     @property bool empty() @safe pure nothrow @nogc
     {
         return pos == max_states;
     }
 
+    ///
     @property CombinationsRepeat save() @safe pure nothrow
     {
         typeof(this) c = this;
--- a/source/mir/combinatorics/package.d
+++ b/source/mir/combinatorics/package.d
@@ -1,0 +1,1109 @@
+/**
+This module contains various combinatorics algorithms.
+
+Authors: Sebastian Wilzbach, Ilya Yaroshenko
+
+License: $(LINK2 http://boost.org/LICENSE_1_0.txt, Boost License 1.0).
+*/
+module mir.combinatorics;
+
+import std.range.primitives: isInputRange, hasLength;
+import std.traits: isFloatingPoint;
+
+/**
+Checks whether we can do basic arithmetic operations, comparisons, modulo and
+assign values to the type.
+*/
+private template isArithmetic(R)
+{
+    enum bool isArithmetic = is(typeof(
+    (inout int = 0)
+    {
+        R r = 1;
+        R test = (r * r / r + r - r) % r;
+        if(r < r && r > r) {}
+    }));
+}
+
+/**
+Checks whether we can do basic arithmetic operations, comparison and modulo
+between two types. R needs to support item assignment of S (it includes S).
+Both R and S need to be arithmetic types themselves.
+*/
+private template isArithmetic(R, S)
+{
+    enum bool isArithmetic = is(typeof(
+    (inout int = 0)
+    {
+        if(isArithmetic!R && isArithmetic!S) {}
+        S s = 1;
+        R r = 1;
+        R test = r * s + r * s;
+        R test2 = r / s + r / s;
+        R test3 = r - s + r - s;
+        R test4 = r % s + r % s;
+        if(r < s && s > r) {}
+        if(s < r && r > s) {}
+    }));
+}
+
+/**
+Computes the $(WEB en.wikipedia.org/wiki/Binomial_coefficient, binomial coefficient)
+of n and k.
+It is also known as "n choose k" or more formally as $(D _n!/_k!(_n-_k)).
+If a fixed-length integer type is used and an overflow happens, `0` is returned.
+
+Uses the generalized binomial coefficient for negative integers and floating
+point number
+
+Params:
+    n = arbitrary arithmetic type
+    k = arbitrary arithmetic type
+
+Returns:
+    Binomial coefficient
+*/
+
+R binomial(R = ulong, T)(T n, T k)
+    if ((isArithmetic!(R, T) || isFloatingPoint!R) &&
+        ((is(typeof(T.min < 0)) && is(typeof(T.init & 1))) || !is(typeof(T.min < 0))) )
+{
+    import std.traits: isFloatingPoint;
+    static if (!isFloatingPoint!R)
+    {
+        R result = 1;
+
+        enum hasMinProperty = is(typeof(T.min < 0));
+        // only add negative support if possible
+        static if ((hasMinProperty && T.min < 0) || !hasMinProperty)
+        {
+            if (n < 0)
+            {
+                if (k >= 0)
+                {
+                    return (k & 1 ? -1 : 1) * binomial!(R, T)(-n + k-1, k);
+                }
+                else if (k <= n)
+                {
+                    return ((n-k) & 1 ? -1 : 1) * binomial!(R, T)(-k-1, n-k);
+                }
+            }
+            if (k < 0)
+            {
+                result = 0;
+                return result;
+            }
+        }
+
+        if (k > n)
+        {
+            result = 0;
+            return result;
+        }
+        if (k > n - k)
+        {
+            k = n - k;
+        }
+        // make a copy of n (could be a custom type)
+        for (T i = 1, m = n; i <= k; i++, m--)
+        {
+            // check whether an overflow can happen
+            // hasMember!(Result, "max") doesn't work with dmd2.068 and ldc 0.17
+            static if (is(typeof(0 > R.max)))
+            {
+                if (result / i > R.max / m) return 0;
+                result = result / i * m + result % i * m / i;
+            }
+            else
+            {
+                result = result * m / i;
+            }
+        }
+        return result;
+    }
+    else
+    {
+        import std.math: exp;
+        import std.mathspecial: logGamma;
+        real x = exp(logGamma(n+1)-logGamma(k+1)-logGamma(n-k+1));
+        return x;
+    }
+}
+
+///
+pure @safe unittest
+{
+    assert(binomial(5, 2) == 10);
+    assert(binomial(6, 4) == 15);
+    assert(binomial(3, 1) == 3);
+
+    import std.bigint;
+    assert(binomial!BigInt(1000, 10) == BigInt("263409560461970212832400"));
+}
+
+pure nothrow @safe @nogc unittest
+{
+    assert(binomial(5, 1) == 5);
+    assert(binomial(5, 0) == 1);
+    assert(binomial(1, 2) == 0);
+    assert(binomial(1, 0) == 1);
+    assert(binomial(1, 1) == 1);
+    assert(binomial(2, 1) == 2);
+    assert(binomial(2, 1) == 2);
+
+    // negative
+    assert(binomial!long(-5, 3) == -35);
+    assert(binomial!long(5, -3) == 0);
+}
+
+unittest
+{
+    import std.bigint;
+
+    // test larger numbers
+    assert(binomial(100, 10) == 17310309456440);
+    assert(binomial(999, 5) == 8209039793949);
+    assert(binomial(300, 10) == 1398320233241701770LU);
+    assert(binomial(300LU, 10LU) == 1398320233241701770LU);
+
+    // test overflow
+    assert(binomial(500, 10) == 0);
+
+    // all parameters as custom types
+    BigInt n = 1010, k = 9;
+    assert(binomial!BigInt(n, k) == BigInt("2908077120956865974260"));
+
+    // negative
+    assert(binomial!BigInt(-5, 3) == -35);
+    assert(binomial!BigInt(5, -3) == 0);
+    assert(binomial!BigInt(-5, -7) == 15);
+
+    // floating point
+    import std.math: approxEqual;
+    assert(binomial!real(5.8, 3).approxEqual(17.632, 0.001));
+    assert(binomial!real(20.8, 3.2).approxEqual(1783.34, 0.01));
+    assert(binomial!real(-5.3, 4).approxEqual(84.2958375, 0.0001));
+    assert(binomial!real(-5.6, -3.4).approxEqual(0.00998410, 0.00001));
+    assert(binomial!real(-5.7, -7.6).approxEqual(20.1795, 0.0001));
+
+    // big floating point
+    assert(binomial!real(1.2e10, 2.3).approxEqual(5.66794e22, 0.0001));
+    // TODO: rounding errors seem to sum up here
+    //assert(binomial!real(2.245e15, 5.7).approxEqual(7.68342e84, 0.1));
+}
+
+private struct IndexedRoR(Collection, Range)
+if (isInputRange!Range)
+{
+    import std.range : indexed, isForwardRange;
+
+    private Collection c;
+    private Range r;
+
+    this(Collection collection, Range range)
+    {
+        this.c = collection;
+        this.r = range;
+    }
+
+    @property auto ref front()
+    {
+        return r.indexed(c.front);
+    }
+
+    void popFront()
+    {
+        c.popFront;
+    }
+
+    @property bool empty()
+    {
+        return c.empty;
+    }
+
+    static if(hasLength!Collection)
+    {
+        @property size_t length()
+        {
+            return c.length;
+        }
+    }
+
+    static if(isForwardRange!Collection)
+    {
+        @property typeof(this) save()
+        {
+            return IndexedRoR!(Collection, Range)(c.save, r);
+        }
+    }
+}
+
+/**
+Creates a projection of `collection` Range (e.g. `Combinations`)
+onto a custom `range`.
+*/
+IndexedRoR!(Collection, Range) indexedRoR(Collection, Range)(Collection c, Range r)
+if (isInputRange!Range)
+{
+    return IndexedRoR!(Collection, Range)(c, r);
+}
+
+
+
+/**
+Lazily computes all _permutations of `r` using $(WEB
+en.wikipedia.org/wiki/Heap%27s_algorithm, Heap's algorithm).
+
+While generating a new item is in `O(k)` (amortized `O(1)`),
+the number of permutations is `|n|!`.
+
+Params:
+    n = number of elements (`|r|`)
+    r = $(REF_ALTTEXT Input range, isInputRange, std, range, primitives)
+    alloc = custom Allocator
+
+Returns:
+    Forward range which yields the permutations
+*/
+Permutations permutations(size_t n) @safe pure nothrow
+{
+    assert(n, "must have at least one item");
+    return Permutations(new uint[n-1], new uint[n]);
+}
+
+/// ditto
+IndexedRoR!(Permutations, Range) permutations(Range)(Range r) @safe pure nothrow
+if (isInputRange!Range && hasLength!Range)
+{
+    return permutations(r.length).indexedRoR(r);
+}
+
+///
+pure @safe nothrow unittest
+{
+    import std.algorithm.comparison : equal;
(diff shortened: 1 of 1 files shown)


wilzbach:
As requested at the [phobos PR](https://github.com/D-Programming-Language/phobos/pull/4026) here are `binomial`, `cartesianPower`, `combinations` and `combinationsRepeat`.

1) `bionomial` now supports arbitrary types like `bigint`
2) `bionomial` checks for overflows only if it's possible and returns zero in such a case
3) Logically `permutations` (from std.algorithms.iteration) felt to belong to this category, I copied it over as it wanted to modify it's documentation header and as if this is merged back to phobos will either go back into `iteration` or change `permutations` location.
4) there was a weird bug with dmd2.068 and ldc0.17 - `hasMember!(ulong, "max")` evaluates to false.
5) Complexities: I intend to go through the literature and find the according algortihms

source/mir/combinatorics.d
```diff
@@ -0,0 +1,734 @@
...
+
+Params:
+    n = non negative integer
+    k = non negative integer, k <= n
+
+Returns:
+    Binomial coefficient
+*/
+
+Result binomial(Result, NType, KType)(NType n, KType k)
```
9il: `NType` should the same `KType`, just single `T` for both `n` и `k`.
`Result` should be replaced with `R = typeof(T.init - T.init)`.

source/mir/combinatorics.d
```diff
@@ -0,0 +1,734 @@
...
+        else
+        {
+            result = result * m / i;
+        }
+    }
+    return result;
+}
+
+/// ditto
+size_t binomial(size_t n, size_t k) pure nothrow @safe @nogc
```
9il: This prototype can be removed.

source/mir/combinatorics.d
```diff
@@ -0,0 +1,734 @@
...
+    }
+    // make a copy of n (could be a custom type)
+    for (KType i = 1, m = n; i <= k; i++, m--)
+    {
+        // check whether an overflow can happen
+        // hasMember!(Result, "max") doesn't work with dmd2.068 and ldc 0.17
+        static if (is(typeof((inout int = 0)
+        {
+            if (Result.max) {}
+        })))
```
9il: `is(...) && !isFloatingPoint!R`

source/mir/combinatorics.d
```diff
@@ -0,0 +1,734 @@
...
+    assert(binomial(5, 1) == 5);
+    assert(binomial(5, 0) == 1);
+    assert(binomial(1, 2) == 0);
+    assert(binomial(1, 0) == 1);
+    assert(binomial(1, 1) == 1);
+    assert(binomial(2, 1) == 2);
+    assert(binomial(2, 1) == 2);
+}
+
+unittest
```
9il: Needs test with floating point numbers (including fractional n and k)

source/mir/combinatorics.d
```diff
@@ -0,0 +1,734 @@
...
+                import std.algorithm.mutation : swap;
+
+                if (n > indices.length)
+                {
+                    _empty = true;
+                    return;
+                }
+
+                if (n % 2 == 1)
+                    swap(indices[0], indices[n-1]);
```
9il: Please use `swapAt`

source/mir/combinatorics.d
```diff
@@ -0,0 +1,734 @@
...
+                else
+                    swap(indices[state[n-2]], indices[n-1]);
```
9il: ditto

source/mir/combinatorics.d
9il: ... in UFCS style `indices.swapAt(0, n-1);`

source/mir/combinatorics.d
```diff
@@ -0,0 +1,734 @@
...
+            {
+                import std.algorithm.mutation : swap;
+
+                if (n > indices.length)
+                {
+                    _empty = true;
+                    return;
+                }
+
+                if (n % 2 == 1)
```
9il: `if(n & 1)`

source/mir/combinatorics.d
```diff
@@ -0,0 +1,734 @@
...
+
+While generating a new item is in `O(k)` (amortized `O(1)`),
+the number of permutations is `|r|!`.
+
+Returns:
+A forward range the elements of which are an $(XREF range,
+indexed) view into `r`.
+*/
+auto permutations(Range)(Range r)
+if (isRandomAccessRange!Range && hasLength!Range)
```
9il: Needs an integration with `std.experimental.allocator`

source/mir/combinatorics.d
```diff
@@ -0,0 +1,734 @@
...
+{
+    static struct Permutations
+    {
+        size_t[] indices, state;
```
9il: private?

source/mir/combinatorics.d
```diff
@@ -0,0 +1,734 @@
...
+
+        @property auto front()
+        {
+            import std.range : indexed;
+            return r.indexed(indices);
+        }
+
+        void popFront()
+        {
+            void next(int n)
```
9il: Why this functions is needed?

source/mir/combinatorics.d
```diff
@@ -0,0 +1,734 @@
+module mir.combinatorics;
+
+private import std.range: isRandomAccessRange, hasLength;
```
9il: import by default is private

source/mir/combinatorics.d
```diff
@@ -0,0 +1,734 @@
...
+            import std.range : iota;
+            import std.array : array;
+
+            this.r = r;
+            state = r.length ? new size_t[r.length-1] : null;
+            indices = iota(size_t(r.length)).array;
+            _empty = r.length == 0;
+        }
+
+        private bool _empty;
```
9il: Please move all variable declarations to the beginning of the struct

source/mir/combinatorics.d
9il: Oh, I see. Please move this function out and make it private.

source/mir/combinatorics.d
```diff
@@ -0,0 +1,734 @@
...
+
+            // set initial state and calculate max possibilities
+            if (r.length > 0)
+            {
+                _max_states = binomial(r.length, repeat);
+                _state = new size_t[](repeat);
+                // skip first duplicate
+                if (r.length > 1 && repeat > 1)
+                {
+                    _state = iota(repeat).array;
```
9il: This is not good. First you allocate an array with `new`. Than you allocate another with `.array`. Please check other such cases.

source/mir/combinatorics.d
9il: change module to `std.range.primitives`

source/mir/combinatorics.d
9il: Better options: looks like this is just simple loop `for(size_t h = 0;;h++) { ... }` where `return` replaced with `break` and `h` is `n-2` (`h` would be a little bit faster).

source/mir/combinatorics.d
9il: `if(h & 1)`

source/mir/combinatorics.d
9il: `indices.swapAt(0, h+1);`

source/mir/combinatorics.d
```diff
@@ -0,0 +1,734 @@
...
+
+                if (n % 2 == 1)
+                    swap(indices[0], indices[n-1]);
+                else
+                    swap(indices[state[n-2]], indices[n-1]);
+
+                if (++state[n-2] == n)
+                {
+                    state[n-2] = 0;
+                    next(n+1);
```
9il: just add `break` here

source/mir/combinatorics.d
```diff
@@ -0,0 +1,734 @@
...
+                }
```
9il:
```
if (state[h] == h+1)
{
    state[h] = 0;
    continue;
}
state[h]++;
break;
```

9il:
Something go wrong with LDC

wilzbach:
1) Wow - I just copied `permutations` over without looking deeply at it Phobos. I will remember to more critical to code from Phobos in the future.

2) swapAt is [not documentated](https://github.com/D-Programming-Language/phobos/blob/master/std/algorithm/mutation.d#L2004) - how about a PR at phobos?

3) I added support for negative number and floating point number

4) Floating point binomial runs into rounding issues for large n - e.g. [SciPy](https://github.com/scipy/scipy/blob/81c096001974f0b5efe29ec83b54f725cc681540/scipy/special/orthogonal_eval.pxd#L109) uses a more sophisticated algorithm. I will check if I find litrature about that, s.t. we can use something similar

5) Some addressed points in summary

> Result should be replaced with R = typeof(T.init - T.init).

ehm wouldn't this automatically result in the same type as T?
I prefer a different strategy - use the largest unsigned type (ulong) by default and let the user specify a different type if wished - this covers the following cases nicely

``` D
// n and k are low numbers, the result is big
binomial!BigInt(1000, 10);
// n and k are by default int -> an overflow just happened
bionomial(100, 10);
```

> Needs test with floating point numbers (including fractional n and k)

what do you mean by this?
I added some tests for floating point numbers, but afaik there is no rational type yet in phobos

6) I added some constraints to the binomial coefficient, they might be a bit complicated, but what they basically say is that if there is a type that supports all these arithmetic operations, we can compute the binomial coefficient. The last line is there because bit comparison is only needed for signed types. The check for `isFloatingPoint` is optional (all floating types are already included because they support rudimentary arithemetic operations, however it makes the header easier to read.

7) ldc weirdness - probably this has something to do with the precision of real. I guess as soon as 4) is fixed, ldc will stop to complain.
ldc returns 4.47729e+85

8) will work on integration with the experimental allocator soon - do you like the way it is done for [EMSI containers](https://github.com/economicmodeling/containers)?

9il:
> 2) swapAt is not documentated - how about a PR at phobos?

Yes

> 5) Some addressed points in summary
> 
> Result should be replaced with R = typeof(T.init - T.init).
> ehm wouldn't this automatically result in the same type as T?

No, type of `byte - byte` is int. May be we ca use only `uint` and FP types?

wilzbach:
@9il I added an experimental interface to the allocator API.
It does support adding a custom allocator now and even `@nogc` works :)
- it stops being `pure` (that's why I still use `new` for the default GCAllocator)
- it expects the allocator type to have a `.instance` property - I am not sure whether this is a fair assumption, but it simplifies our code a lot.. 
- only if version is `> 2.069` (hence a few conditionals in the code).

What do you think about this design?

> No, type of byte - byte is int. May be we ca use only uint and FP types?

Negative binomial is defined too. What is your opinion about my current solution with the `isArithmetic` template and having `ulong` as default target type that can be overwritten if needed?

9il:
Please drop support for dmd FE <= 2.068 for this module.

---

No all allocators has `instance`. Not all are shared.

``` D
(Allocator, Range)(Range r, Allocator alloc)
```

and wrapper for `GCAllocator`:

``` D
(Range)(Range r)
{
// import GCAllocator internal
}
```

9il:
> isArithmetic

Good idea to reduce imports, thanks!

wilzbach:
> Please drop support for dmd FE <= 2.068 for this module.

I would like to, but the allocator was introduced in 2.070, so even LDC would fail.
Btw that's also why I didn't use local GCAllocator import's yet, because we still need the stub.

> No all allocators has instance.

I updated `combinationsRepeat`, making Allocator optional has a couple of caveats:
- we have to use two methods instead of one (that's why I made the struct public)
- we have to have two different constructors (as GCAllocator.instance is already impure - we cannot access mutable static data 'instance')

> Not all are shared.

At least GCAllocator needs to be `shared` - is there any way we can infer this?

See `combinationsRepeat`:

https://github.com/DlangScience/mir/pull/29/files#diff-def635b35c404dc25be3ab3c785e833fR791

source/mir/combinatorics/package.d
```diff
@@ -0,0 +1,1006 @@
...
+*/
+auto combinationsRepeat(Allocator, Range)(Range r, size_t k, Allocator alloc)
+if (isRandomAccessRange!Range && hasLength!Range)
+in
+{
+    assert(k >= 1, "Invalid number of combinations");
+}
+body
+{
+
```
9il: blankline

source/mir/combinatorics/package.d
```diff
@@ -0,0 +1,1006 @@
...
+        if (r.length > 0)
+        {
+            _max_states = binomial(r.length + repeat  - 1, repeat);
+            _state = new size_t[](repeat);
+        }
+    }
+
+    static if(!is(Allocator == GCAllocator))
+    {
+        this(Range r, size_t repeat, shared Allocator alloc)
```
9il: remove `shared`

source/mir/combinatorics/package.d
```diff
@@ -0,0 +1,1006 @@
...
+/// ditto
+auto combinationsRepeat(Range)(Range r, size_t k = 1)
+if (isRandomAccessRange!Range && hasLength!Range)
+in
+{
+    assert(k >= 1, "Invalid number of combinations");
+}
+body
+{
+    return CombinationRepeat!(GCAllocator, Range)(r, k);
```
9il: need import for `GCAllocator`

source/mir/combinatorics/package.d
```diff
@@ -0,0 +1,1006 @@
...
+License: $(LINK2 http://boost.org/LICENSE_1_0.txt, Boost License 1.0).
+*/
+module mir.combinatorics;
+
+import std.range.primitives: isRandomAccessRange, hasLength;
+import std.traits: isFloatingPoint;
+
+static if (__VERSION__ > 2069)
+{
+    import std.experimental.allocator.gc_allocator: GCAllocator;
```
9il: should be removed from global

source/mir/combinatorics/package.d
```diff
@@ -0,0 +1,1006 @@
...
+            {
+                _max_states = pow(r.length, repeat);
+                static if(is(Allocator == GCAllocator))
+                    _state = new size_t[](repeat);
+                else
+                    _state = alloc.makeArray!size_t(repeat);
+            }
+        }
+
+        static if(!is(Allocator == GCAllocator))
```
9il: ~~Probably this can solved by user~~

9il: hmm, lets use this code, thought

9il:
> At least GCAllocator needs to be shared - is there any way we can infer this?

No worries. `shared` would be a part of type in template difinition

source/mir/combinatorics/package.d
```diff
@@ -0,0 +1,1006 @@
...
+        {
+            import std.range : iota;
+            import std.array : array;
+
+            this.r = r;
+            alloc = alloc;
+
+            size_t rLength = r.length;
+
+            static if(is(Allocator == GCAllocator))
```
9il: This type of conditions can be removed, just use common case :)

source/mir/combinatorics/package.d
```diff
@@ -0,0 +1,1006 @@
...
+        static if (__VERSION__ > 2069)
+            import std.experimental.allocator: makeArray;
+        private size_t[] indices, state;
+        private Range r;
+        private bool _empty;
+        shared Allocator alloc;
+
+        this(Range r, shared Allocator alloc)
+        {
+            import std.range : iota;
```
9il: Try to avoid imports if possible

source/mir/combinatorics/package.d
wilzbach: I would like too, but then we loose the purity of the struct and function.

9il: OK, lets keep the code

source/mir/combinatorics/package.d
```diff
@@ -0,0 +1,987 @@
...
+    assert(binomial!real(5.8, 3).approxEqual(17.632, 0.001));
+    assert(binomial!real(20.8, 3.2).approxEqual(1783.34, 0.01));
+    assert(binomial!real(-5.3, 4).approxEqual(84.2958375, 0.0001));
+    assert(binomial!real(-5.6, -3.4).approxEqual(0.00998410, 0.00001));
+    assert(binomial!real(-5.7, -7.6).approxEqual(20.1795, 0.0001));
+
+    // big floating point
+    assert(binomial!real(1.2e10, 2.3).approxEqual(5.66794e22, 0.0001));
+    // TODO: rounding errors seem to sum up here
+    //assert(binomial!real(2.245e15, 5.7).approxEqual(7.68342e84, 0.1));
```
wilzbach: this might be a future problem, when using binomial with floats/reals

source/mir/combinatorics/package.d
```diff
@@ -0,0 +1,987 @@
...
+
+While generating a new item is in `O(k)` (amortized `O(1)`),
+the number of permutations is `|r|!`.
+
+Returns:
+A forward range the elements of which are an $(XREF range,
+indexed) view into `r`.
+*/
+auto permutations(Allocator = bool, Range)(Range r,
+                  Allocator alloc = Allocator.init)
```
wilzbach: using `bool` as default type is a hack, s.t. that we have just one constructor and function. It makes the constructor easier. Is there a better default type that can be used a parameter?

The alternative is to put the struct out as `private struct` as done before.

9il: Oh, don't do this kind of hack please)

wilzbach: You prefer a `private struct` or is there a better way?

9il: Why you don't want to have two prototypes ?

9il: private struct is OK

wilzbach: because there is still some shared part of the constructor - ok will move that part into a own function and put everything to the struct ;-)
Thanks for your input!

9il: > because there is still some shared

???

9il: Ah, understand now)

wilzbach: already fixed - don't know why github is still showing this.

source/mir/combinatorics/package.d
```diff
@@ -0,0 +1,1089 @@
...
+en.wikipedia.org/wiki/Heap%27s_algorithm, Heap's algorithm).
+
+While generating a new item is in `O(k)` (amortized `O(1)`),
+the number of permutations is `|r|!`.
+
+Returns:
+A forward range the elements of which are an $(XREF range,
+indexed) view into `r`.
+*/
+auto permutations(Range, Allocator)(Range r, auto ref Allocator alloc)
```
wilzbach: Inspired by your allocate API PR:
- I changed the order, so that the type can automatically be found
- `alloc` is passed as `ref` and only stored as a pointer

source/mir/combinatorics/package.d
```diff
@@ -0,0 +1,1089 @@
...
+    private bool _empty;
+
+    enum useGC = is(Allocator == void);
+    static if(!useGC)
+        Allocator* alloc;
+
+    private void _init()
+    {
+        _empty = r.length == 0;
+    }
```
wilzbach: I could put this into both constructor, but for other structs (below) it's more logic - so I went for a common behavior.

source/mir/combinatorics/package.d
```diff
@@ -0,0 +1,1089 @@
...
+}
+
+/// ditto
+auto permutations(Range)(Range r)
+if (isRandomAccessRange!Range && hasLength!Range)
+{
+    return Permutations!Range(r);
+}
+
+private struct Permutations(Range, Allocator = void)
```
wilzbach: I needed a way to pass the Allocator optionally to the struct - as GCAllocator is not currently available, I used `void`. This time it is _hidden_ from the user and I actually like it a lot more than the `GCAllocator` stub.

wilzbach:
@9il - I reworked the allocation support and moved all nested the structs out.
This time I am actually happy with the result. Thanks you so much for your helpful & enormous feedback that help{ed,s} me to learn.

As this is getting in its final shape - I squashed all commits into one.

source/mir/combinatorics/package.d
9il: `private` should be removed

9il: `Allocator` should be removed too

9il: just construct this struct inside a function

source/mir/combinatorics/package.d
9il: ???

9il: default value would be 0

source/mir/combinatorics/package.d
wilzbach: How? I have to save the reference to the Allocator somehow and afaik it isn't possible to extend a struct.

source/mir/combinatorics/package.d
wilzbach: The other constructors share some common bits - I put them into such a init function to avoid duplication.

source/mir/combinatorics/package.d
```diff
@@ -0,0 +1,1090 @@
...
+    enum useGC = is(Allocator == void);
+    static if(!useGC)
+        Allocator* alloc;
+
+    private void _init()
+    {
+        _empty = r.length == 0;
+    }
+
+    this(Range r)
```
9il: should accept indexes and state arrays

source/mir/combinatorics/package.d
9il: add a `dispose` method for user

source/mir/combinatorics/package.d
```diff
@@ -0,0 +1,1090 @@
...
+            foreach(i, ref el; indices)
+            {
+                el = iotaResult[i];
+            }
+            alloc = alloc;
+
+            _init();
+        }
+
+        ~this()
```
9il: just remove this

source/mir/combinatorics/package.d
```diff
@@ -0,0 +1,1090 @@
...
+        this.r = r;
+        size_t rLength = r.length;
+
+        state = new size_t[rLength - 1];
+        indices = iota(size_t(rLength)).array;
+
+        _init();
+    }
+
+    static if(!useGC)
```
9il: move to the `make ***` function

source/mir/combinatorics/package.d
wilzbach: Okay that looks nicer - thanks!

```
    this(Allocator)(Range r, ref Allocator alloc)
    {
        ...
    }

    void dispose(Allocator)(Allocator alloc)
    {
       ...
    }
```

But isn't it pretty inconvenient (and easy to forgot) for an user to dispose the structs internal state values manually?
AFAICT as soon as the structs gets destructed, they are not needed anymore. 

Moreover the other `structs` support `save`, which also needs allocation, thus saving the reference to the allocation is needed imho.

source/mir/combinatorics/package.d
wilzbach: the other structs support the typical `save` (and now permutations too) - is this what you are asking for?

source/mir/combinatorics/package.d
wilzbach: Update: Compare [permutations](https://github.com/DlangScience/mir/pull/29/files#diff-def635b35c404dc25be3ab3c785e833fR203) (not saving the reference to the allocator) with [cartesianPower](https://github.com/DlangScience/mir/pull/29/files#diff-def635b35c404dc25be3ab3c785e833fR403).
The current permutation test also doesn't call dispose (yet).

source/mir/combinatorics/package.d
wilzbach: Do you mean I should add a special `make` function for allocation?

source/mir/combinatorics/package.d
wilzbach: Another point for storing the Allocator ref:

> It is assumed the respective entities had been allocated with the same allocator.

(from the [dispose api docs](https://dlang.org/library/std/experimental/allocator/dispose.html))

source/mir/combinatorics/package.d
9il: Oh, my bad. Have not seen `==`

source/mir/combinatorics/package.d
9il:
``` D
/// UPDATED
this(Range r, size_t[] indices, size_t[] state)
in{
    size_t rLength = r.length;
    assert(state.length = rLength - 1);
    assert(indices.length = rLength);
}
body {
    foreach(i, ref index; indices)
        index = i; //less imports
    state[] = 0;
    this.indices = indices;
    this.state = state;
    this.r = r;
}
```

source/mir/combinatorics/package.d
```diff
@@ -0,0 +1,1136 @@
...
+        this.r = r;
+        size_t rLength = r.length;
+
+        state = new size_t[rLength - 1];
+        indices = iota(size_t(rLength)).array;
+
+        _init();
+    }
+
+    this(Allocator)(Range r, ref Allocator alloc)
```
9il: please move this logic to `makePermutations`

source/mir/combinatorics/package.d
```diff
@@ -0,0 +1,1136 @@
...
+        foreach(i, ref el; indices)
+        {
+            el = iotaResult[i];
+        }
+        alloc = alloc;
+
+        _init();
+    }
+
+    void dispose(Allocator)(Allocator alloc)
```
9il: please move this outside of the struct

source/mir/combinatorics/package.d
```diff
@@ -0,0 +1,1136 @@
...
+        return c;
+    }
+
+    @property typeof(this) save(Allocator)(ref Allocator alloc)
+    {
+        import std.experimental.allocator: makeArray;
+        typeof(this) c = this;
+        size_t rLength = indices.length;
+
+        c.indices = alloc.makeArray!size_t(rLength);
```
9il: `c.indices = alloc.makeArray!size_t(indices);`

source/mir/combinatorics/package.d
```diff
@@ -0,0 +1,1136 @@
...
+        foreach(i, r; indices)
+            c.indices[i] = indices[i];
+
+        c.state = alloc.makeArray!size_t(rLength - 1);
```
9il: `c.state = alloc.makeArray!size_t(state);`

source/mir/combinatorics/package.d
```diff
@@ -0,0 +1,1136 @@
...
+
+    @property typeof(this) save()
+    {
+        typeof(this) c = this;
+        c.indices = indices.dup;
+        c.state = state.dup;
+        return c;
+    }
+
+    @property typeof(this) save(Allocator)(ref Allocator alloc)
```
9il: move this outside the struct

source/mir/combinatorics/package.d
9il: `indices` and `state` should be allocated in `permutations` / `makePermutations`

source/mir/combinatorics/package.d
wilzbach: `save` is the standard method in Phobos to save the state of a struct / function. Imho it should work when using an Allocator too - otherwise a lot of the range algorithms will not work. Most of them expect a `ForwardRange`.
https://dlang.org/library/std/range/interfaces/forward_range.html

9il: i asked only about `save` with allocator. Ok lets preserve both in struct

source/mir/combinatorics/package.d
wilzbach: I am afraid I don't understand why you would expose these two arrays to the user?
Those are internal state variables that are likely to change. A future version of permutations might only need `state` or use more allocated variables.

An alternative would be to use an allocated `struct` which contains this data like

```
PermutationsAllocation permsAllocated = makePermutations(Allocator alloc, 4); // we need k for the length
auto perms = permutations(4, permsAllocated); // weneed k again
dispose(permsAllocated, alloc);
```

In any case I also don't understand why you want to force the user to call `makePermutations` and `dispose` himself. Do you mind to explain this?

The most popular package that I know of that uses the allcoator API are the gcfree [containers](https://github.com/economicmodeling/containers). They save the allocator too.

source/mir/combinatorics/package.d
wilzbach: Ok I guess this is really about whether we should keep a reference (or copy) to the allocator in our struct or not. If we keep the reference.

`save` is called without any variable in those algorithms, hence I would make `static if` variants of `save` depending on whether an Allocator was passed - otherwise I guess moving `save` out makes sense.

9il: Algorithms can be extended with save with allocators ;)

source/mir/combinatorics/package.d
9il: > In any case I also don't understand why you want to force the user to call makePermutations and dispose himself. Do you mind to explain this?

I don't force user to do something, I just want to have more control. Anyway a user will be able to use GC version without any additional thoughts.

9il:
After thinking about API and algorithms I would like to suggest to remove all templated code (except `make***` functions) and use Ranges of indexes instead of templates.
This allows:
1. Reduce code size
2. Allows user to use same indexes with different ranges without ugly `zip` magic
3. Maintain more simple code

In the same time user will be able to achieve the same result as with template using `perm.map!(p => userRange.indexed(p))`.

source/mir/combinatorics/package.d
9il: > Algorithms can be extended with save with allocators ;)

hmm, this was bad idea

9il:
> After thinking about API and algorithms I would like to suggest to remove all templated code (except make**\* functions) and use Ranges of indexes instead of templates.
> This allows:
> 1. Reduce code size
> 2. Allows user to use same indexes with different ranges without ugly zip magic
> 3. Maintain more simple code
> 
> In the same time user will be able to achieve the same result as with template using perm.map!(p => userRange.indexed(p)).

What do you think?

wilzbach:
> What do you think?

I am currently traveling thus the delayed answer,but afaik one still has to do the allocation, do you don't gain anything, but make it more complex.

9il:
> I am currently traveling thus the delayed answer,but afaik one still has to do the allocation, do you don't gain anything, but make it more complex.
>  Show all checks

Allocation should be done in `permutations` and `makePermutations`. Do you see any problems with it?

9il:
BTW, would you mind me to merge this and open PR with changes?

wilzbach:
> BTW, would you mind me to merge this and open PR with changes?

Not at all. Should I squash the commits?

9il:
> Should I squash the commits?

When this would be ready.

So, you don't want to separate logic and templates?

wilzbach:
> So, you don't want to separate logic and templates?

My first choice would be to use just the allocation api, but this leads to an impure struct, but yeah I do prefer having a convenient api for the user and this means to combine the logic and templates.

9il:
> combine the logic and templates

because you want to store allocator's pointer?

My point about convenient api is simple: if user wants convenient and simple API he will use GC. If user wants to use allocators he need to think about allocation logic.

9il:
Note, that this is not a container. This is a simple permutation range

wilzbach:
> Note, that this is not a container. This is a simple permutation range

Exactly my point that I am trying to make. In 99% of the use cases a user wants just to iterate through the range and once he his finished and leaves the scope, he doesn't need it anymore. Thus  automatic deallocation should happen. 

Btw I would be very nice of the language would have more built in support for the allocation api..

9il:
> In 99% of the use cases

... user will use GC.

9il:
Allocators support is for libraries developers, not for a common user. And libraries developers needs full control with minimal effort.

wilzbach:
Okay - do you want to post a suggestion how you imagine the api to look like? I guess we are a lot faster this way.

Update: allocation api

9il:
``` D
Permutations permutations(size_t n) @safe // and etc
{
    assert(n); //???
    return Permutations(new size_t[n-1], new size_t[n]);
}

unittest
{
    import std.algorithm.iteration: indexed; //?
    auto rp = permutations(r.length).map!(perm => r.indexed(perm));
    //...
}

Permutations makePermutations(Allocator)(auto ref Allocator alloc, size_t n)
{
    import std.experimental.allocator: makeArray;
    //...
}


void dispose(Allocator)(auto ref Allocator alloc, auto ref Permutations perm)
{
    import std.experimental.allocator: dispose;
    //...
}

struct Permutations //Just a struct
{
    this(size_t[] indices, size_t[] state) @safe // and etc
    in{
        assert(state.length + 1 = indices.length);
    }
    body {
        foreach(i, ref index; indices)
            index = i; //less imports
        state[] = 0;
        this.indices = indices;
        this.state = state;
    }
}
```

source/mir/combinatorics/package.d
```diff
@@ -947,7 +945,7 @@ public:
         immutable repeat = state.length;
 
         // behaves like: do _getNextState();  while(!_state.isSorted);
-        size_t i = repeat - 1;
+        uint i = cast(uint) repeat - 1;
```
9il: why `i` is uint?

source/mir/combinatorics/package.d
```diff
@@ -961,7 +959,7 @@ public:
         // e.g. [0, 2] -> (upper block: [1, 2]) -> [1, 1]
         if (i != repeat - 1)
         {
-            for (size_t j = i + 1; j < repeat; j++)
+            for (uint j = i + 1; j < repeat; j++)
```
9il: why `j` is uint? why not simple foreach loop?

source/mir/combinatorics/package.d
```diff
@@ -915,13 +913,13 @@ private struct CombinationsRepeat
...
-    size_t[] state;
-    size_t n;
+    uint[] state;
+    uint n;
 
     size_t max_states, pos;
 
 public:
-    this(size_t n, size_t[] state) @safe pure nothrow @nogc
+    this(uint n, uint[] state) @safe pure nothrow @nogc
```
9il: use `size_t` here

9il: .. for `n`

source/mir/combinatorics/package.d
```diff
@@ -915,13 +913,13 @@ private struct CombinationsRepeat
...
     {
         this.n = n;
```
9il: but cast here

source/mir/combinatorics/package.d
```diff
@@ -254,10 +254,10 @@ Returns:
 A forward range the elements of which are an $(XREF range,
 indexed) view into `r`.
 */
-Permutations permutations(size_t n) @safe pure nothrow
+Permutations permutations(uint n) @safe pure nothrow
```
9il: ~~make cast inside the functiton~~

9il: Why you use uint anywhere????
We need it only for internal data storage !

source/mir/combinatorics/package.d
```diff
@@ -301,12 +301,12 @@ pure @safe nothrow unittest
 }
 
 /// ditto
-Permutations makePermutations(Allocator)(auto ref Allocator alloc, size_t n)
+Permutations makePermutations(Allocator)(auto ref Allocator alloc, uint n)
```
9il: ditto
`size_t n`

source/mir/combinatorics/package.d
wilzbach: why? - it will be called via the wrappers and they have `cast` k too

wilzbach:
> make cast inside the functiton

[done](https://github.com/DlangScience/mir/pull/29/commits/9c076b608bcb5a8a1e51abbfce52463289004630)

source/mir/combinatorics/package.d
```diff
@@ -257,7 +257,8 @@ indexed) view into `r`.
 Permutations permutations(size_t n) @safe pure nothrow
 {
     assert(n, "must have at least one item");
-    return Permutations(new size_t[n-1], new size_t[n]);
+    uint nu = cast(uint) n;
```
9il: Why do you need cast here???????????

source/mir/combinatorics/package.d
```diff
@@ -427,7 +427,7 @@ in
 }
 body
 {
-    return CartesianPower(n, new size_t[repeat]);
+    return CartesianPower(cast(uint) n, new uint[cast(uint) repeat]);
```
9il: please move cast to the constructor

source/mir/combinatorics/package.d
```diff
@@ -451,7 +451,7 @@ in
 body
 {
     import std.experimental.allocator: makeArray;
-    return CartesianPower(n, alloc.makeArray!size_t(repeat));
+    return CartesianPower(cast(uint) n, alloc.makeArray!uint(cast(uint) repeat));
```
9il: ditto

source/mir/combinatorics/package.d
```diff
@@ -613,11 +613,11 @@ in
 }
 body
 {
-    return Combinations(n, new size_t[k]);
+    return Combinations(cast(uint) n, new uint[cast(uint) k]);
```
9il: ditto

source/mir/combinatorics/package.d
```diff
@@ -651,17 +651,17 @@ private struct Combinations
...
-    size_t[] state;
-    size_t n;
+    uint[] state;
+    uint n;
     size_t max_states, pos;
 
 public:
 
-    this(size_t n, size_t[] state) @safe pure nothrow @nogc
+    this(uint n, uint[] state) @safe pure nothrow @nogc
```
9il: ditto

source/mir/combinatorics/package.d
9il: ditto

source/mir/combinatorics/package.d
wilzbach: because `uint`[size_t n]` doesn't work

source/mir/combinatorics/package.d
wilzbach: I need to cast `repeat` - as uint[size_t repeat]` is invalid - therefore casting at the same place made sense to me

source/mir/combinatorics/package.d
9il: it works

source/mir/combinatorics/package.d
9il: > uint[size_t repeat]` is invalid

this is not true

wilzbach:
> > uint[size_t repeat]` is invalid
> > this is not true

Oh I am terribly sorry - that was my naive understanding then :<
I am impressed how much I can do wrong. Thanks for your patience!

wilzbach:
image

Seems like `/// ditto` isn't working that well?

9il:
> Seems like /// ditto isn't working that well?

ddox is not very well

wilzbach:
well I guess we have to switch to a better documentation engine soon :/

image

btw what do you think about naming the module `comb` or `combs`?
(I like short names)

9il:
> btw what do you think about naming the module comb or combs?

`combinatorics` looks clear, imho

wilzbach:
> combinatorics looks clear, imho

Okay :)

Anything else you want to see in the initial version?

source/mir/combinatorics/package.d
```diff
@@ -0,0 +1,1109 @@
...
+
+Returns:
+    Binomial coefficient
+*/
+
+R binomial(R = ulong, T)(T n, T k)
+    if ((isArithmetic!(R, T) || isFloatingPoint!R) &&
+        ((is(typeof(T.min < 0)) && is(typeof(T.init & 1))) || !is(typeof(T.min < 0))) )
+{
+    import std.traits: isFloatingPoint;
```
9il: can be removed

source/mir/combinatorics/package.d
```diff
@@ -0,0 +1,1109 @@
...
+    // is copyable?
+    import std.algorithm: equal;
+    import std.range: iota, dropOne;
+    auto a = iota(2).permutations;
+    assert(a.front.equal([0, 1]));
+    assert(a.save.dropOne.front.equal([1, 0]));
+    assert(a.front.equal([0, 1]));
+}
+
+/// ditto
```
9il, line 301: This should be documented

source/mir/combinatorics/package.d
```diff
@@ -0,0 +1,1109 @@
...
+            index = i;
+        state[] = 0;
+
+        this.indices = indices;
+        this.state = state;
+
+        _empty = indices.length == 0;
+    }
+
+    @property uint[] front() @safe pure nothrow @nogc
```
9il: Document with empty `///`

9il: and other range primitives

source/mir/combinatorics/package.d
```diff
@@ -0,0 +1,1109 @@
...
+        {
+            return IndexedRoR!(Collection, Range)(c.save, r);
+        }
+    }
+}
+
+/**
+Creates a projection of `collection` Range (e.g. `Combinations`)
+onto a custom `range`.
+*/
```
9il, line 231: Params and Returns for all public functions

source/mir/combinatorics/package.d
```diff
@@ -0,0 +1,1109 @@
...
+While generating a new item is in `O(k)` (amortized `O(1)`),
+the number of permutations is `|n|!`.
+
+Params:
+    n = number of elements (`|r|`)
+    r = $(REF_ALTTEXT Input range, isInputRange, std, range, primitives)
+    alloc = custom Allocator
+
+Returns:
+    Forward range which yields the permutations
```
9il: `Forward range, which ...` if i am not wrong

source/mir/combinatorics/package.d
```diff
@@ -0,0 +1,1109 @@
...
+}
+
+/// Dispose a CartesianPower object
+void dispose(Allocator)(auto ref Allocator alloc, auto ref CartesianPower cartesianPower)
+{
+    import std.experimental.allocator: dispose;
+    dispose(alloc, cartesianPower._state);
+}
+
+private struct CartesianPower
```
9il: BTW, this struct should be private/package. Just undocumented. Otherwise user would not be able to access methods

source/mir/combinatorics/package.d
wilzbach, line 301: does `void` need a return type?

wilzbach:
Comments addressed in d63ff57. Thanks for the 27th review :) :)

source/mir/combinatorics/package.d
9il, line 301: no

source/mir/combinatorics/package.d
```diff
@@ -0,0 +1,1150 @@
...
+    import std.experimental.allocator.mallocator;
+
+    static immutable expected2 = [[0, 1], [1, 0]];
+    auto r = iota(2);
+    auto rp = makePermutations(Mallocator.instance, r.length);
+    assert(rp.indexedRoR(r).equal!equal(expected2));
+    dispose(Mallocator.instance, rp);
+}
+
+/// ditto
```
9il: What /// ditto?

source/mir/combinatorics/package.d
```diff
@@ -0,0 +1,1150 @@
...
+        state[] = 0;
+
+        this.indices = indices;
+        this.state = state;
+
+        _empty = indices.length == 0;
+    }
+
+    ///
+    @property uint[] front() @safe pure nothrow @nogc
```
9il: `const(uint)[]` and for other structs too.
Also the note about that this range would be changed by the struct after `popFront`, for other structs too

wilzbach: > Also the note about that this range would be changed by the struct after popFront, for other structs too

What do you mean by that?

wilzbach: > const(uint)[] and for other structs too.

addressed.

9il: > Also the note about that this range would be changed by the struct after popFront, for other structs too
> What do you mean by that?

Oh, my English ... :/

We need to note a user that popFront will change content of the returned range

source/mir/combinatorics/package.d
```diff
@@ -0,0 +1,1149 @@
...
+        state[] = 0;
+
+        this.indices = indices;
+        this.state = state;
+
+        _empty = indices.length == 0;
+    }
+
+    ///
+    @property const(uint[]) front() @safe pure nothrow @nogc
```
9il: `const(uint)[]`, but not const(uint[])

source/mir/combinatorics/package.d
wilzbach: > Oh, my English ... :/

No worries - that's how i understood it.

> We need to note a user that popFront will change content of the returned range

Isn't that expected for ranges and popFront?
That's why there are forward ranges that one can save?

source/mir/combinatorics/package.d
wilzbach: sorry - typo :/

source/mir/combinatorics/package.d
9il:
```
const(uint)[] data = p.front;
p.popFront; // `data` is changed here
```

wilzbach: I see - thanks :)
But modifying the indexes doesn't make sense, so why don't we make it entirely `const`?

9il: > But modifying the indexes doesn't make sense, so why don't we make it entirely const?

To disallow a user to break our indices

wilzbach: > To disallow a user to break our indices

Oh I was proposing `const uint[]` - as far as i understand this implies constness for its values.

wilzbach:
no rebase?

9il:
O_o

wilzbach:
> O_o

I am sorry - I tried to avoid total rebases in the end, so that reviewing the changes is easier for you.
Do you want to do a `git rebase` and push force?

9il:
> I am sorry - I tried to avoid total rebases in the end, so that reviewing the changes is easier for you.
> Do you want to do a git rebase and push force?

No, they are squashed

wilzbach:
Ah wow - my fault. Didn't see that.

Btw about the `const` discussion - how about changing `const(uint)[] front()` to `const uint[] front()`?
(Then the user can't change neither the list nor the elements)

9il:
this is bad api

wilzbach:
> this is bad api

why? didn't you want to add as a notice that the user should be careful when modifying the array itself?

9il:
He can change array but not content

wilzbach:
> He can change array but not content

Yep but it is the internal state array, so changing it will lead most likely to an error. 
For a normal user it is proxied via the IndexedRoR struct, so he won't be affected by this constraint?

9il:
Yes

wilzbach:
> Yes

So why is using `const uint[]` a bad API then? Sorry that I still don't understand your reasoning.

9il:
> So why is using const uint[] a bad API then? Sorry that I still don't understand your reasoning.

``` D
auto p = permutations(4);
auto p = p.front;
p = new uint[3];// fails if `const(uint[])`
```

wilzbach:
Thanks for the example and sorry for bothering you!

PR #42, opened 2016-04-10
add iotaSlice

9il:


9il:
ping @wilzbach, this is fix for #22

source/mir/ndslice/selection.d
```diff
@@ -1556,18 +1557,28 @@ IndexSlice!N indexSlice(size_t N)(auto ref size_t[N] lengths)
 }
 
 ///
-@safe @nogc pure nothrow unittest
+@safe pure nothrow unittest
```
wilzbach: I start to dislike  the Allocator now, but it might be worth to think about adding support for it?

9il: What do you mean? `makeSlice` and `dispose` was added to `ndslice.slice`. `indexSlice` is lazy and `@nogc`.

wilzbach: > indexSlice is lazy and @nogc.

You removed the check for nogc here.

9il: Yes, because `[[[0, 0], [0, 1], [0, 2]],  [[1, 0], [1, 1], [1, 2]]]` allocates

wilzbach: `static immutable`?

9il: :-( ok

source/mir/ndslice/selection.d
```diff
@@ -1534,6 +1534,7 @@ pure nothrow unittest
 
 /++
 Returns a slice, the elements of which are equal to the initial index value.
+This is multidimensional analog of $(LINK2 std_range.html#iota, std.range.iota).
```
wilzbach, line 1537: Maybe you could add a see also to iota slice? 
Moreover I think that if you at the multidimensional analog of iota, one might think about iotaSlice, so maybe pointing out that the index of reseted for every dimension could help to solve the confusion? 
And then "for a continuous index, see iotaSlice"

wilzbach:
> ping @wilzbach, this is fix for #22

Looks great - exactly what #22 was about :)

9il:
Thank you for the review!

PR #52, opened 2016-04-10
fix build

9il:


wilzbach:
Probably we should have something like auotester that only merges once CI is passing...

9il:
Yeah)

wilzbach:
Seems like github already supports this 
https://help.github.com/articles/enabling-required-status-checks/

wilzbach:
Did it work?

9il:
yes, thanks!

PR #61, opened 2016-04-10
ndslice.internal: improve coverage

wilzbach:
let's improve mir's coverage - part II

### DIFF
--- a/source/mir/ndslice/internal.d
+++ b/source/mir/ndslice/internal.d
@@ -187,6 +187,20 @@ bool isPermutation(size_t N)(auto ref in size_t[N] perm)
     return true;
 }
 
+unittest
+{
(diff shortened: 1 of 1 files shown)


wilzbach:
if you look at `isPermutation` - the for loop seems unnecessary.
AFAICT the 
- mask has the length of the input numbers 
- the impl checks for duplicates (via the mask)
- the impl checks for numbers which are greater or equal than N (negative is not possible)

So if we have n numbers - they are all either unique (=is_permutation) or a) have duplicates or b) numbers outside the range. Have a look at the lines above the PR:

```
foreach (e; mask)
         if (e == false)
             return false; // never gets executed
```

9il:
rebase please

9il:
oh, do not need

PR #62, opened 2016-04-10
ndslice.slice: improve coverage

wilzbach:
The `index` and `save` methods here are never called.

This might be a bit controversial - so I separated it.

### DIFF
--- a/source/mir/ndslice/slice.d
+++ b/source/mir/ndslice/slice.d
@@ -290,12 +290,15 @@ pure nothrow @nogc unittest
     }
 
     alias S = Slice!(3, MyIota);
-    auto slice = MyIota().sliced(20, 10);
 
     import std.range.primitives;
     static assert(hasLength!S);
     static assert(isInputRange!S);
     static assert(isForwardRange!S == false);
+
+    auto slice = MyIota().sliced(20, 10);
+    assert(slice[1, 2] == 12);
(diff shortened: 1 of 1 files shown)


source/mir/ndslice/slice.d
```diff
@@ -316,14 +317,17 @@ pure nothrow @nogc unittest
...
     static assert(hasLength!S);
     static assert(hasSlicing!S);
     static assert(isForwardRange!S);
     static assert(isBidirectionalRange!S);
     static assert(isRandomAccessRange!S);
+
+    auto slice = MyIota().sliced(20, 10);
+    assert(slice[0, 0] == 0);
+    auto sCopy = slice.save;
+    assert(sCopy[0, 0] == 0);
```
9il: please remove both assert. They are don't help a user at all

9il: and my assert too

wilzbach: Hmm the problem is that the `struct` counts to the coverage as miss :/
Other ideas?

9il: OK, lets get `assert(sCopy[1, 2] == 12);`

wilzbach: done - however we need to call `save` too ...

PR #65, opened 2016-04-11
mir.combinatorics: try external via r.indexed

wilzbach:
Following up on #29 - here's the code for `permutations` when using `r.indexed`.

The biggest problem that I discovered is that closures (aka delegates) are allcated via the gc. I tried to work around this problem by using a struct in which the range is saved. However without success yet.

Convenience wrapper for `Range` are provided

This is a development PR. It contains some dev output, so don't merge. Real work will happen back at #29.

The current output is:

```
[p [0, 1]
r.length 835760380
[-835727616, -835727615], p [1, 0]
r.length 835760380
[-835727615, -835727616]]
```

Just passing the method doesn't save it's context - is there a way to save the context without a delegate?

### DIFF
--- a/source/mir/combinatorics/package.d
+++ b/source/mir/combinatorics/package.d
@@ -1,0 +1,201 @@
+module mir.combinatorics;
+
+import std.range.primitives: isInputRange, hasLength;
+
+struct IndexedRoR(Collection, Range)
+if (isInputRange!Range)
+{
+    import std.range : indexed, isForwardRange;
+
+    private Collection c;
+    private Range r;
+
+    this(Collection c, Range r)
+    {
+        this.c = c;
+        this.r = r;
+    }
+
+    @property auto ref front()
+    {
+        return r.indexed(c.front);
+    }
+
+    void popFront()
+    {
+        c.popFront;
+    }
+
+    @property bool empty()
+    {
+        return c.empty;
+    }
+
+    static if(hasLength!Collection)
+    {
+        @property size_t length()
+        {
+            return c.length;
+        }
+    }
+
+    static if(isForwardRange!Collection && isForwardRange!Range)
+    {
+        @property typeof(this) save()
+        {
+            return IndexedRoR!(Collection, Range)(c.save, r.save);
+        }
+    }
+}
+
+IndexedRoR!(Collection, Range) indexedRoR(Collection, Range)(Collection c, Range r)
+if (isInputRange!Range)
+{
+    return IndexedRoR!(Collection, Range)(c, r);
+}
+
+Permutations permutations(size_t n) @safe pure nothrow
+{
+    assert(n, "must have at least one item");
+    return Permutations(new size_t[n-1], new size_t[n]);
+}
+
+auto permutations(Range)(Range r) @safe pure nothrow
+if (isInputRange!Range && hasLength!Range)
+{
+    auto perms = .permutations(r.length);
+    return IndexedRoR!(Permutations, Range)(perms, r);
+}
+
+///
+pure @safe nothrow unittest
+{
+    import std.algorithm.comparison : equal;
+    import std.range : iota;
+
+    auto expectedRes = [[0, 1, 2],
+         [1, 0, 2],
+         [2, 0, 1],
+         [0, 2, 1],
+         [1, 2, 0],
+         [2, 1, 0]];
+
+    auto r = iota(3);
+    auto rp = permutations(r.length).indexedRoR(r);
+    assert(rp.equal!equal(expectedRes));
+
+    // direct style
+    auto rp2 = iota(3).permutations;
+    assert(rp2.equal!equal(expectedRes));
+}
+
+pure @safe nothrow unittest
+{
+    // is copyable?
+    import std.algorithm: equal;
+    import std.range: iota, dropOne;
+    auto a = iota(2).permutations;
+    assert(a.front.equal([0, 1]));
+    assert(a.save.dropOne.front.equal([1, 0]));
+    assert(a.front.equal([0, 1]));
+}
+
+Permutations makePermutations(Allocator)(auto ref Allocator alloc, size_t n)
+{
+    assert(n, "must have at least one item");
+    import std.experimental.allocator: makeArray;
+    auto state = alloc.makeArray!size_t(n - 1);
+    auto indices = alloc.makeArray!size_t(n);
+    return Permutations(state, indices);
+}
+
+void dispose(Allocator)(auto ref Allocator alloc, auto ref Permutations perm)
+{
+    import std.experimental.allocator: dispose;
+    dispose(alloc, perm.state);
+    dispose(alloc, perm.indices);
+}
+
+static if (__VERSION__ > 2069) @nogc unittest
+{
+    import std.algorithm: equal;
+    import std.range : iota;
+
+    import std.experimental.allocator.mallocator;
+
+    static immutable expected2 = [[0, 1], [1, 0]];
+    auto r = iota(2);
+    auto rp = makePermutations(Mallocator.instance, r.length);
+    assert(rp.indexedRoR(r).equal!equal(expected2));
+    dispose(Mallocator.instance, rp);
+}
+
+struct Permutations
+{
+    private size_t[] indices, state;
+    private bool _empty;
+
+    this(size_t[] state, size_t[] indices) @safe pure nothrow @nogc
+    in
+    {
+        assert(state.length + 1 == indices.length);
+    }
+    body
+    {
+        // iota
+        foreach(i, ref index; indices)
+            index = i;
+        state[] = 0;
+
+        this.indices = indices;
+        this.state = state;
+
+        size_t indicesLength = indices.length;
+
+        _empty = indicesLength == 0;
+    }
+
+    @property auto front()
+    {
+        return indices;
+    }
+
+    void popFront() @safe pure nothrow @nogc
+    {
+        import std.algorithm.mutation : swapAt;
+        for (size_t h = 0;;h++)
+        {
+            if (h+2 > indices.length)
+            {
+                _empty = true;
+                break;
+            }
+
+            if (h & 1)
+                indices.swapAt(0, h+1);
+            else
+                indices.swapAt(state[h], h+1);
+
+            if (state[h] == h+1)
+            {
+                state[h] = 0;
+                continue;
+            }
+            state[h]++;
+            break;
+        }
+    }
+
+    @property bool empty() @safe pure nothrow @nogc
+    {
+        return _empty;
+    }
+
+    @property typeof(this) save() @safe pure nothrow
+    {
+        typeof(this) c = this;
+        c.indices = indices.dup;
+        c.state = state.dup;
+        return c;
+    }
+}


source/mir/combinatorics/package.d
```diff
@@ -0,0 +1,184 @@
...
+    import std.algorithm: map;
+    import std.range : indexed, ElementType;
+
+    size_t n = r.length;
+    auto state = alloc.makeArray!size_t(n - 1);
+    auto indices = alloc.makeArray!size_t(n);
+    auto perms = Permutations(state, indices);
+
+    auto mapper = Mapper!Range(r);
+    auto d = &mapper.mapper;
```
wilzbach: this is the problematic part

source/mir/combinatorics/package.d
```diff
@@ -0,0 +1,184 @@
...
+}
+
+auto permutations(Range)(Range r) @safe pure nothrow
+if (isRandomAccessRange!Range && hasLength!Range)
+{
+    import std.range : indexed;
+    import std.algorithm: map;
+    size_t n = r.length;
+    assert(n); //???
+    return Permutations(new size_t[n-1], new size_t[n]).map!(p => r.indexed(p));
```
9il: call `. permutations` internally.

source/mir/combinatorics/package.d
```diff
@@ -0,0 +1,184 @@
...
+if (isRandomAccessRange!Range && hasLength!Range)
+{
+    import std.experimental.allocator: makeArray;
+    import std.algorithm: map;
+    import std.range : indexed, ElementType;
+
+    size_t n = r.length;
+    auto state = alloc.makeArray!size_t(n - 1);
+    auto indices = alloc.makeArray!size_t(n);
+    auto perms = Permutations(state, indices);
```
9il: call `. makePermutations` internally.

source/mir/combinatorics/package.d
9il: BTW, can we just remove `(Range)` template? This one like of code

source/mir/combinatorics/package.d
9il: > BTW, can we just remove (Range) template? This one like of code

ditto

source/mir/combinatorics/package.d
```diff
@@ -0,0 +1,184 @@
...
+
+void dispose(Allocator)(auto ref Allocator alloc, auto ref Permutations perm)
+{
+    import std.experimental.allocator: dispose;
+    dispose(alloc, perm.state);
+    dispose(alloc, perm.indices);
+}
+
+static if (__VERSION__ > 2069)
+{
```
9il: Can remove braces for `static if`

source/mir/combinatorics/package.d
9il: Just move this to unittest

wilzbach: `permutations` is the only method that is partially copied from Phobos -
https://dlang.org/library/std/algorithm/iteration/permutations.html

So I guess we have to provide a Range interface.

9il: I know, but Phobos is not the bible

source/mir/combinatorics/package.d
```diff
@@ -0,0 +1,182 @@
+module mir.combinatorics;
+
+import std.range.primitives: isRandomAccessRange, hasLength;
+import std.stdio;
+
+Permutations permutations(size_t n) @safe pure nothrow
+{
+    return Permutations(new size_t[n-1], new size_t[n]);
```
9il, line 60: `assert(n);`

source/mir/combinatorics/package.d
wilzbach: The main problem is that can't use delegates in nogc and if we find a workaround it will probably difficult for the user too?

wilzbach:
@9il - sorry if I didn't point this out in a better way. The reason why I pushed this is because I couldn't figure out a way how to use `map` in the allocation function. I tried a workaround with the external struct, but even this failed.

I know that the other parts are not perfect, but I saw that the `indexed` approach probably isn't going to work, so this was thought to be the proof.

source/mir/combinatorics/package.d
9il: Yes, but we can provide a singl struct like `IndexedRoR`, which could work will any struct for `combinatorics`

9il:
> @9il - sorry if I didn't point this out in a better way. The reason why I pushed this is because I couldn't figure out a way how to use map in the allocation function. I tried a workaround with the external struct, but even this failed.
> 
> I know that the other parts are not perfect, but I saw that the indexed approach probably isn't going to work, so this was thought to be the proof.
>  Hide all checks

So, we can provide single `IndexedRoR` struct for delegates workaround.

source/mir/combinatorics/package.d
```diff
@@ -0,0 +1,373 @@
...
+        c.popFront;
+    }
+
+    @property bool empty()
+    {
+        return c.empty;
+    }
+
+    @property size_t length()
+    {
```
wilzbach, line 35: it might happen that people try to create ranges with a greater length.

source/mir/combinatorics/package.d
```diff
@@ -0,0 +1,373 @@
...
+
+    dispose(Mallocator.instance, rp);
+}
+}
+
+struct Permutations
+{
+    private size_t[] indices, state;
+    private bool _empty;
+    size_t _max_states, _pos;
```
wilzbach: limits us to state.length of 20 on x64

source/mir/combinatorics/package.d
```diff
@@ -0,0 +1,373 @@
...
+    @property size_t length()
+    {
+        return c.length;
+    }
+
+    static if(isForwardRange!Range)
+    {
+        @property typeof(this) save()
+        {
+            typeof(this) t = this;
```
9il:
```
typeof(this) t = this(c.save, r.save);
// ... copy other fields.
```

this would be faster

9il: fixed

source/mir/combinatorics/package.d
wilzbach: now we have two `@nogc` tests

source/mir/combinatorics/package.d
```diff
@@ -0,0 +1,393 @@
...
+    }
+
+    @property typeof(this) save(Allocator)(auto ref Allocator alloc)
+    {
+        typeof(this) p = makePermutations(alloc, indices.length);
+        p.state[] = state;
+        p.indices[] = indices;
+        p._max_states = _max_states;
+        p._pos = _pos;
+        p._empty = _empty;
```
wilzbach: the other way with `this` somehow didn't work.

source/mir/combinatorics/package.d
```diff
@@ -0,0 +1,393 @@
...
+    {
+        @property typeof(this) save()
+        {
+            typeof(this) t = this;
+            t.c = c.save;
+            t.r = r.save;
+            return t;
+        }
+
+        @property typeof(this) save(Allocator)(auto ref Allocator alloc)
```
9il: Just remove this for now. API is not clrear

source/mir/combinatorics/package.d
```diff
@@ -0,0 +1,393 @@
...
+
+    @property typeof(this) save() @safe pure nothrow
+    {
+        typeof(this) c = this;
+        c.indices = indices.dup;
+        c.state = state.dup;
+        return c;
+    }
+
+    @property typeof(this) save(Allocator)(auto ref Allocator alloc)
```
9il: remove this please too

source/mir/combinatorics/package.d
wilzbach: > fixed

?

source/mir/combinatorics/package.d
```diff
@@ -0,0 +1,393 @@
...
+equal n. Its generalization to real numbers is called the Gamma function.
+As the factorial grows very quickly, it is recommended to use `BigInt`.
+
+Params:
+    n = factorial number
+
+Returns:
+    Computed factorial
+*/
+auto factorial(R, T)(T n)
```
9il: I don't think we need this functions. Only factorial for BigInts is interesting and in the same time proper algorithm is complicated

source/mir/combinatorics/package.d
9il: > fixed
> ?

```
typeof(this) t = this(c.save, r.save);
// ... copy other fields.
```

source/mir/combinatorics/package.d
wilzbach: It is needed to compute the length of the permutation range

9il: Just use `foreach` =)

source/mir/combinatorics/package.d
wilzbach: this gives `constructor call must be in a constructor` - but we can call IndexedRoR ;-)

source/mir/combinatorics/package.d
```diff
@@ -0,0 +1,201 @@
...
+
+IndexedRoR!(Collection, Range) indexedRoR(Collection, Range)(Collection c, Range r)
+if (isInputRange!Range)
+{
+    return IndexedRoR!(Collection, Range)(c, r);
+}
+
+Permutations permutations(size_t n) @safe pure nothrow
+{
+    return Permutations(new size_t[n-1], new size_t[n]);
```
9il, line 60: `assert(n)` please

source/mir/combinatorics/package.d
```diff
@@ -0,0 +1,201 @@
...
+    import std.range: iota, dropOne;
+    auto a = iota(2).permutations;
+    assert(a.front.equal([0, 1]));
+    assert(a.save.dropOne.front.equal([1, 0]));
+    assert(a.front.equal([0, 1]));
+}
+
+Permutations makePermutations(Allocator)(auto ref Allocator alloc, size_t n)
+{
+    import std.experimental.allocator: makeArray;
```
9il, line 106: `assert(n)`

wilzbach:
> So, we can provide single IndexedRoR struct for delegates workaround.

this was a great idea! Should I start to convert the other functions with a similar style?

9il:
Yes, thank you)

9il:
BTW, would `state` and `indexes` be always OK with uint? If so, please change it. This will reduce memory usage and cache misses.

source/mir/combinatorics/package.d
```diff
@@ -0,0 +1,201 @@
...
+
+        this.indices = indices;
+        this.state = state;
+
+        size_t indicesLength = indices.length;
+
+        _empty = indicesLength == 0;
+    }
+
+    @property auto front()
```
9il, line 158: please put type for documentation `uint[] front() @safe pure nothrow @nogc`

source/mir/combinatorics/package.d
```diff
@@ -0,0 +1,201 @@
...
+    static immutable expected2 = [[0, 1], [1, 0]];
+    auto r = iota(2);
+    auto rp = makePermutations(Mallocator.instance, r.length);
+    assert(rp.indexedRoR(r).equal!equal(expected2));
+    dispose(Mallocator.instance, rp);
+}
+
+struct Permutations
+{
+    private size_t[] indices, state;
```
9il, line 135: looks like they ok to be `uint`

wilzbach:
> BTW, would state and indexes be always OK with uint?

ubyte should be fine too - 256! has ~10^506 possibilities.
http://www.wolframalpha.com/input/?i=256!

9il:
> ubyte should be fine too - 256! has ~10^506 possibilities.

ushort and ubyte are slow

source/mir/combinatorics/package.d
wilzbach, line 135: ehm we use a lot of `.length` and sometimes `index i` which are both type `size_t`, thus we need to do casting - should i still use uint?

9il, line 135: yes

source/mir/combinatorics/package.d
```diff
@@ -0,0 +1,201 @@
...
+
+    this(size_t[] state, size_t[] indices) @safe pure nothrow @nogc
+    in
+    {
+        assert(state.length + 1 == indices.length);
+    }
+    body
+    {
+        // iota
+        foreach(i, ref index; indices)
```
9il, line 146: foreach(uint i, ref index; indices)

source/mir/combinatorics/package.d
```diff
@@ -0,0 +1,201 @@
...
+            index = i;
+        state[] = 0;
+
+        this.indices = indices;
+        this.state = state;
+
+        size_t indicesLength = indices.length;
+
+        _empty = indicesLength == 0;
```
9il, line 155: empty = indices.length == 0;

source/mir/combinatorics/package.d
wilzbach, line 135: FYI I had to use `cast` about 6x times 
https://github.com/DlangScience/mir/pull/29/commits/46039f4dececa165eb0dfd8d6adb9e6b316f0569

wilzbach:
Closing as #29 has been rebased and changed according to this idea.

PR #74, opened 2016-04-11
Update README.md

9il:


wilzbach:
how about adding "analog to Python's combinatoric generators from itertools"?

9il:
No, they are simple anought

PR #82, opened 2016-04-13
fast byElement

wilzbach:
Hey I think I read your warning once that `byElement` shouldn't be used if RandomAccess is not needed.

Well I was wondering whether there is a better way to iterate element-wise through a slice and I didn't find anything, so I quickly wrote it myself.

It is so simple, that I expect this functionality to exist somewhere - so I didn't clean the code well.
In case this is really new, let me know and I am happy to invest more time.

### DIFF
--- a/source/mir/ndslice/selection.d
+++ b/source/mir/ndslice/selection.d
@@ -1697,3 +1697,111 @@ struct IotaMap
         return index;
     }
 }
+
+/**
+Element-wise iteration over a slice.
+In contrast to byElement, it only returns a bidirectional forward range access.
+However it is faster for such operations.
+*/
+auto byElementIterator(size_t N, Range)(auto ref Slice!(N, Range) slice)
+{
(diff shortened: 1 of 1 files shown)


source/mir/ndslice/selection.d
```diff
@@ -1697,3 +1697,111 @@ struct IotaMap
         return index;
     }
 }
+
+/**
+Element-wise iteration over a slice.
+In contrast to byElement, it only returns a bidirectional forward range access.
+However it is faster for such operations.
+*/
+auto byElementIterator(size_t N, Range)(auto ref Slice!(N, Range) slice)
```
9il, line 1706: This works only for completely dense slices in normal form. This will not work after iteration operators. If slice has completely dense normal form byElement can return an array.

9il:
Dont spend time for that right now ) We have more important things todo

PR #85, opened 2016-04-13
mir.random: initial API

wilzbach:
Hey,

so I started to think (and play) with a new random API.
Please don't invest time looking over the code - it's just a draft and I want to use to ask better questions.
Update: only have a [lookat the unittest at the bottom](https://github.com/DlangScience/mir/pull/85/files#diff-a1906e86b965bd2806454874420b9c70R137)

Questions
1) Do we want methods to use a `struct` or just a `function` for the random generator methods? 
If you have a peek at [Boost::Random](https://github.com/boostorg/random), you will see that [`reset`](https://github.com/boostorg/random/blob/develop/include/boost/random/gamma_distribution.hpp#L155) is quite common. However [dstats](https://github.com/dsimcha/dstats/blob/master/source/dstats/random.d) avoids general states, but stores them if needed as a [singleton](https://github.com/dsimcha/dstats/blob/master/source/dstats/random.d#L485).

2) Do we want to save a reference to the random generator in such a struct?
The advantage is that when passing the reference, the struct is an usual infinite forward Range (see 3 too).

3) If the struct gets duplicates via `save` - should the random generator get duplicated too? A la

```
assert(randomMethod.front.approxEqual(0.786));
assert(randomMethod.save.dropOne.front.approxEqual(0.831));
assert(randomMethod.front.approxEqual(0.786));
```

4) The names are still horrible, but do you prefer to manual construction of a `struct` like `rarray(Bernoulli(0.5), 10)` or having a wrapper around it `rarray!Bernoulli(10, 0.5)` or both styles?

### DIFF
--- a/source/mir/random/package.d
+++ b/source/mir/random/package.d
@@ -1,0 +1,182 @@
+/**
+Random number generators.
+
+Authors: Sebastian Wilzbach, Ilya Yaroshenko
+
+License: $(LINK2 http://boost.org/LICENSE_1_0.txt, Boost License 1.0).
+*/
+module mir.random;
+
+import std.random: rndGen, Random, uniform;
+import mir.ndslice.slice: Slice;
+
+/*
+Checks whether a type is Slice.
+/**/
+template isSlice(T)
+{
+    import std.traits: TemplateOf;
+    enum bool isSlice = __traits(compiles, __traits(isSame, TemplateOf!(T), Slice));
+}
+
+// init distribution with random method and args
+auto rinit(alias Random, T, Args...)(T arr, Args args)
+if (!isSlice!T)
+{
+    foreach (ref el; arr)
+    {
+        el = Random(args);
+    }
+    return arr;
+}
+
+// init slice distribution
+auto rinit(alias Random, size_t N, Range, Args...)(auto ref Slice!(N, Range) slice, Args args)
+{
+    import mir.ndslice.selection: byElement;
+    auto il = byElement(slice);
+    rinit!Random(il, args);
+    return slice;
+}
+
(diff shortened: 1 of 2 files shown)


9il:
1. No range interfaces

Function should accept RNG and return value

9il:
... and state if required

source/mir/random/package.d
```diff
@@ -0,0 +1,182 @@
...
+
+/*
+WIP - Currently only
+$(WEB https://en.wikipedia.org/wiki/Box%E2%80%93Muller_transform, Box Muller-transform)
+References:
+    Butcher, J. C. "Random sampling from the normal distribution."
+    The Computer Journal 3.4 (1961): 251-253.
+    http://comjnl.oxfordjournals.org/content/3/4/251.full.pdf
+*/
+auto normal(RGen)(double mu, double sigma, ref RGen gen = rndGen)
```
9il, line 147: we need compression of different normal RNGs

9il, line 147: I found 3 API errors in this function... we just need good common C style functions. API is the last step for this module

wilzbach, line 147: do you mean compression or comparison?

> we just need good common C style functions. API is the last step for this module

ok!

9il, line 147: > I found 3 API errors in this function... we just need good common C style functions. API is the last step for this module

comparison))) There are different normal RNG and we should implement the best or few best algorithms

9il:
1. structs
2. yes
3. no
4. see mir.random.algorithm in mir-random

Should be reopened for mir-random

PR #86, opened 2016-04-13
add template isSlice in ndslice.slice

wilzbach:
As a Slice can be a Range, I was finding the template `isSlice` quite useful when working on the random API. Probably it makes sense to expose this template to the user too - what would be the best spot to add it?

### DIFF
--- a/source/mir/ndslice/slice.d
+++ b/source/mir/ndslice/slice.d
@@ -23,6 +23,29 @@ import std.range.primitives; //: hasLength;
 
 import mir.ndslice.internal;
 
+/*
+Checks whether a type is Slice.
+**/
+template isSlice(T)
+{
+    import std.traits: TemplateOf;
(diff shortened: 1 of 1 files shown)


9il:
Lets think about it after sparse matrixes

wilzbach:
> Lets think about it after sparse matrixes

ok, but they will have a different type?
E.g. `isSparse` and maybe unified to `isMatrix`?

9il:
> ok, but they will have a different type?

This is complex question

wilzbach:
update: it's already part of slide.d, but just `private`

``` D
private enum bool _isSlice(T) = is(T : Slice!(N, Range), size_t N, Range);
```

see https://github.com/DlangScience/mir/blob/master/source/mir/ndslice/slice.d#L140

9il:
> update: it's already part of slide.d, but just private

OK, please move this declaration to `ndslice.slice` and make it public

9il:
Should be opened in discussed in Phobos

PR #88, opened 2016-04-14
add linting via dscanner

wilzbach:
I think we should try putting the linting off to a real bot.
This PR adds a execution of [dscanner](https://github.com/Hackerpilot/Dscanner).

We probably want to play with this for a while, maybe disable some checks or fix the codebase until we set the `--enforce` flag, which will exit the CI with an error.
Currently it outputs many messages, see e.g. [here](https://travis-ci.org/wilzbach/mir/jobs/122970079)

Other
- It is only run on travis subbuild - hence the version check (I reused it for the s3 check)
- As there is no binary release of dscanner, I created `releases.dlang.io` (a s3 bucket to which I will upload such release) - this speeds up the build! Otherwise there is still dub fetch & dub run.

### DIFF
--- a/source/mir/combinatorics/package.d
+++ b/source/mir/combinatorics/package.d
@@ -7,7 +7,18 @@ License: $(LINK2 http://boost.org/LICENSE_1_0.txt, Boost License 1.0).
 */
 module mir.combinatorics;
 
-import std.range.primitives: isInputRange, hasLength;
+import std.range.primitives: isRandomAccessRange, hasLength;
+
+///
+unittest
+{
+    import std.algorithm.comparison: equal;
+
+    assert([0, 1].permutations.equal!equal([[0, 1], [1, 0]]));
+    assert([0, 1].cartesianPower(2).equal!equal([[0, 0], [0, 1], [1, 0], [1, 1]]));
+    assert([0, 1].combinations(2).equal!equal([[0, 1]]));
+    assert([0, 1].combinationsRepeat(2).equal!equal([[0, 0], [0, 1], [1, 1]]));
+}
 
 /**
 Checks whether we can do basic arithmetic operations, comparisons, modulo and
@@ -49,7 +60,7 @@ private template isArithmetic(R, S)
 /**
 Computes the $(WEB en.wikipedia.org/wiki/Binomial_coefficient, binomial coefficient)
 of n and k.
-It is also known as "n choose k" or more formally as $(D _n!/_k!(_n-_k)).
+It is also known as "n choose k" or more formally as `_n!/_k!(_n-_k)`.
 If a fixed-length integer type is used and an overflow happens, `0` is returned.
 
 Uses the generalized binomial coefficient for negative integers and floating
@@ -62,7 +73,6 @@ Params:
 Returns:
     Binomial coefficient
 */
-
 R binomial(R = ulong, T)(T n, T k)
     if (isArithmetic!(R, T) &&
         ((is(typeof(T.min < 0)) && is(typeof(T.init & 1))) || !is(typeof(T.min < 0))) )
@@ -149,10 +159,10 @@ unittest
     import std.bigint;
 
     // test larger numbers
-    assert(binomial(100, 10) == 17310309456440);
-    assert(binomial(999, 5) == 8209039793949);
-    assert(binomial(300, 10) == 1398320233241701770LU);
-    assert(binomial(300LU, 10LU) == 1398320233241701770LU);
+    assert(binomial(100, 10) == 17_310_309_456_440);
+    assert(binomial(999, 5) == 82_09_039_793_949);
+    assert(binomial(300, 10) == 1_398_320_233_241_701_770LU);
+    assert(binomial(300LU, 10LU) == 1_398_320_233_241_701_770LU);
 
     // test overflow
     assert(binomial(500, 10) == 0);
@@ -167,14 +177,37 @@ unittest
     assert(binomial!BigInt(-5, -7) == 15);
 }
 
+/**
+Creates a projection of a generalized `Collection` range for the numeric case
+case starting from `0` onto a custom `range` of any type.
+
+Params:
+    collection = range to be projected from
+    range = random access range to be projected to
+
+Returns:
+    Range with a projection to range for every element of collection
+
+See_Also:
+    $(LREF permutations), $(LREF cartesianPower), $(LREF combinations),
+    $(LREF combinationsRepeat)
+*/
+IndexedRoR!(Collection, Range) indexedRoR(Collection, Range)(Collection c, Range r)
+if (isRandomAccessRange!Range)
+{
+    return IndexedRoR!(Collection, Range)(c, r);
+}
+
+/// ditto
 struct IndexedRoR(Collection, Range)
-if (isInputRange!Range)
+if (isRandomAccessRange!Range)
 {
     import std.range : indexed, isForwardRange;
 
     private Collection c;
     private Range r;
 
+    ///
     this(Collection collection, Range range)
     {
         this.c = collection;
@@ -182,7 +215,7 @@ if (isInputRange!Range)
     }
 
     ///
-    @property auto ref front()
+    auto ref front() @property
     {
         return r.indexed(c.front);
     }
@@ -194,7 +227,7 @@ if (isInputRange!Range)
     }
 
     ///
-    @property bool empty()
+    bool empty() @property const
     {
         return c.empty;
     }
@@ -202,7 +235,7 @@ if (isInputRange!Range)
     static if(hasLength!Collection)
     {
         ///
-        @property size_t length()
+        @property size_t length() const
         {
             return c.length;
         }
@@ -211,31 +244,62 @@ if (isInputRange!Range)
     static if(isForwardRange!Collection)
     {
         ///
-        @property typeof(this) save()
+        typeof(this) save() @property
         {
             return IndexedRoR!(Collection, Range)(c.save, r);
         }
     }
 }
 
-/**
-Creates a projection of `collection` Range (e.g. `Combinations`)
-onto a custom `range`.
-
-Params:
-    collenction = Range to be projected from
-    range = Range to be projected to
-
-Returns:
-    Range with a projection to range for every element of collection
-*/
-IndexedRoR!(Collection, Range) indexedRoR(Collection, Range)(Collection c, Range r)
-if (isInputRange!Range)
+///
+@safe pure nothrow unittest
 {
-    return IndexedRoR!(Collection, Range)(c, r);
+    import std.algorithm.comparison: equal;
+
+    auto perms = 2.permutations;
+    assert(perms.save.equal!equal([[0, 1], [1, 0]]));
+
+    auto projection = perms.indexedRoR([1, 2]);
+    assert(projection.equal!equal([[1, 2], [2, 1]]));
 }
 
+///
+unittest
+{
+    import std.algorithm.comparison: equal;
+    import std.range: only;
 
+    auto projectionD = 2.permutations.indexedRoR("ab"d);
+    assert(projectionD.equal!equal([['a', 'b'], ['b', 'a']]));
+
+    auto projectionC = 2.permutations.indexedRoR(only('a', 'b'));
+    assert(projectionC.equal!equal([['a', 'b'], ['b', 'a']]));
+}
+
+@safe pure nothrow unittest
+{
+    import std.algorithm.comparison: equal;
+    import std.range: dropOne;
+
+    auto perms = 2.permutations;
+    auto projection = perms.indexedRoR([1, 2]);
+    assert(projection.length == 2);
+
+    // can save
+    assert(projection.save.dropOne.front.equal([2, 1]));
+    assert(projection.front.equal([1, 2]));
+}
+
+@safe nothrow @nogc unittest
+{
+    import std.algorithm.comparison: equal;
+    static perms = 2.permutations;
+    static immutable projectionArray = [1, 2];
+    auto projection = perms.indexedRoR(projectionArray);
+
+    static immutable result = [[1, 2], [2, 1]];
+    assert(projection.equal!equal(result));
+}
 
 /**
 Lazily computes all _permutations of `r` using $(WEB
@@ -246,11 +310,14 @@ the number of permutations is `|n|!`.
 
 Params:
     n = number of elements (`|r|`)
-    r = $(REF_ALTTEXT Input range, isInputRange, std, range, primitives)
+    r = $(REF_ALTTEXT Random access range, isRandomAccessRange, std, range, primitives)
     alloc = custom Allocator
 
 Returns:
     Forward range, which yields the permutations
+
+See_Also:
+    $(LREF Permutations)
 */
 Permutations permutations(size_t n) @safe pure nothrow
 {
@@ -260,42 +327,9 @@ Permutations permutations(size_t n) @safe pure nothrow
 
 /// ditto
 IndexedRoR!(Permutations, Range) permutations(Range)(Range r) @safe pure nothrow
-if (isInputRange!Range && hasLength!Range)
+if (isRandomAccessRange!Range)
 {
     return permutations(r.length).indexedRoR(r);
-}
-
-///
-pure @safe nothrow unittest
-{
-    import std.algorithm.comparison : equal;
-    import std.range : iota;
-
-    auto expectedRes = [[0, 1, 2],
-         [1, 0, 2],
-         [2, 0, 1],
-         [0, 2, 1],
-         [1, 2, 0],
-         [2, 1, 0]];
-
-    auto r = iota(3);
-    auto rp = permutations(r.length).indexedRoR(r);
-    assert(rp.equal!equal(expectedRes));
-
-    // direct style
-    auto rp2 = iota(3).permutations;
-    assert(rp2.equal!equal(expectedRes));
-}
-
-pure @safe nothrow unittest
-{
-    // is copyable?
-    import std.algorithm: equal;
-    import std.range: iota, dropOne;
-    auto a = iota(2).permutations;
-    assert(a.front.equal([0, 1]));
-    assert(a.save.dropOne.front.equal([1, 0]));
-    assert(a.front.equal([0, 1]));
 }
 
 /// ditto
@@ -308,41 +342,29 @@ Permutations makePermutations(Allocator)(auto ref Allocator alloc, size_t n)
     return Permutations(state, indices);
 }
 
-/*
-Disposes a Permutations object. It destroys and then deallocates the
-Permutations object pointed to by a pointer.
-It is assumed the respective entities had been allocated with the same allocator.
+/**
+Lazy Forward range of permutations using $(WEB
+en.wikipedia.org/wiki/Heap%27s_algorithm, Heap's algorithm).
 
-Params:
-    alloc = Custom allocator
-    perm = Permutations object
+It always generates the permutations from 0 to `n - 1`,
+use $(LREF indexedRoR) to map it to your range.
+
+Generating a new item is in `O(k)` (amortized `O(1)`),
+the total number of elements is `n^k`.
+
+See_Also:
+    $(LREF permutations), $(LREF makePermutations)
 */
-void dispose(Allocator)(auto ref Allocator alloc, auto ref Permutations perm)
-{
-    import std.experimental.allocator: dispose;
-    dispose(alloc, perm.state);
-    dispose(alloc, perm.indices);
-}
-
-static if (__VERSION__ > 2069) @nogc unittest
-{
-    import std.algorithm: equal;
-    import std.range : iota;
-
-    import std.experimental.allocator.mallocator;
-
-    static immutable expected2 = [[0, 1], [1, 0]];
-    auto r = iota(2);
-    auto rp = makePermutations(Mallocator.instance, r.length);
-    assert(rp.indexedRoR(r).equal!equal(expected2));
-    dispose(Mallocator.instance, rp);
-}
-
 struct Permutations
 {
     private uint[] indices, state;
     private bool _empty;
+    size_t _max_states = 1, _pos;
 
+    /**
+    state should have the length of `n - 1`,
+    whereas the length of indices should be `n`
+    */
     this(uint[] state, uint[] indices) @safe pure nothrow @nogc
     in
     {
@@ -359,6 +381,10 @@ struct Permutations
         this.state = state;
 
         _empty = indices.length == 0;
+
+        // factorial
+        foreach (i; 1..indices.length + 1)
+            _max_states *= i;
     }
 
     ///
@@ -371,6 +397,10 @@ struct Permutations
     void popFront() @safe pure nothrow @nogc
     {
         import std.algorithm.mutation : swapAt;
+
+        assert(!empty);
+        _pos++;
+
         for (uint h = 0;;h++)
         {
             if (h+2 > indices.length)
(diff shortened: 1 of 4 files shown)


9il:
I don't think dscanner is good enought

wilzbach:
> I don't think dscanner is good enought

AFAIK it's the best tool out there and we can just disable the warnings that we don't like.
In my opinion a linting is better than no linting?

9il:
ping me when you would think PR is ready

wilzbach:
@9il I think I am getting somewhere, but for the remaining issues I would like to hear your opinion and help. I do consider static analysis of code important as it can prevent a couple of errors and it ensures that the codes **sticks to a style guide** - the latter being the main reason for my attempt here.
Dscanner checks for a lot of stuff - have a look at the `.dscanner.ini`. It was really helpful to realize that the new combinatorics module didn't document itself properly.

You can run it locally with:

```
dub fetch dscanner --version=~master
dub run dscanner -- --config .dscanner.ini --styleCheck source
```

Or install the `dscanner-git` package on Arch Linux. I recommend storing the binary somewhere as you avoid the `dub` checking and can pass the spammy error log to dev/null. I use

```
dscanner --config .dscanner.ini --styleCheck source 2> /dev/null
```

I grouped them after their warning types, so that it's easier to look at them.
Please do consider that disabling a check because of one or two edge cases results in we no benefit from this help in the future. I still didn't set the `--enforce` flag, but I plan to do so after we resolved remaining the conflicts below.

## Naming style guide

From Phobos: functions + variables in camelCase.

```
source/mir/las/sum.d(2125:11)[warn]: Function name 'SummationAlgo' does not match style guidelines.
source/mir/las/sum.d(2135:11)[warn]: Function name 'SummationAlgo' does not match style guidelines.
source/mir/ndslice/internal.d(174:21)[warn]: Variable name 'DynamicArrayDimensionsCount' does not match style guidelines.
source/mir/ndslice/internal.d(176:21)[warn]: Variable name 'DynamicArrayDimensionsCount' does not match style guidelines.
source/mir/ndslice/internal.d(256:8)[warn]: Struct name '_Slice' does not match style guidelines.
source/mir/ndslice/slice.d(1042:21)[warn]: Variable name 'PureN' does not match style guidelines.
```

## Subtracting from `.length`

```
source/mir/ndslice/selection.d(221:29)[warn]: Avoid subtracting from '.length' as it may be unsigned.
source/mir/ndslice/selection.d(222:29)[warn]: Avoid subtracting from '.length' as it may be unsigned.
```

## Zero-parameter '@property' function

```
source/mir/ndslice/selection.d(937:22)[warn]: Zero-parameter '@property' function should be marked 'const', 'inout', or 'immutable'.
source/mir/ndslice/selection.d(979:22)[warn]: Zero-parameter '@property' function should be marked 'const', 'inout', or 'immutable'.
source/mir/ndslice/selection.d(1422:22)[warn]: Zero-parameter '@property' function should be marked 'const', 'inout', or 'immutable'.
source/mir/ndslice/slice.d(1327:14)[warn]: Zero-parameter '@property' function should be marked 'const', 'inout', or 'immutable'.
source/mir/ndslice/slice.d(1374:14)[warn]: Zero-parameter '@property' function should be marked 'const', 'inout', or 'immutable'.
source/mir/ndslice/slice.d(2845:14)[warn]: Zero-parameter '@property' function should be marked 'const', 'inout', or 'immutable'.
```

## 'opCmp', 'toHash', 'opEquals', 'opCast', and/or 'toString' should be const

```
source/mir/ndslice/slice.d(1591:10)[warn]: Methods 'opCmp', 'toHash', 'opEquals', 'opCast', and/or 'toString' are non-const.
source/mir/ndslice/slice.d(1609:10)[warn]: Methods 'opCmp', 'toHash', 'opEquals', 'opCast', and/or 'toString' are non-const.
```

## has method 'opEquals', but not 'toHash'

```
source/mir/ndslice/slice.d(1027:8)[warn]: 'Slice' has method 'opEquals', but not 'toHash'.
```

## Avoid use of the comma expression

```
source/mir/ndslice/selection.d(763:66)[warn]: Avoid using the comma expression.
```

## Disabled flags

Moreover I disabled a couple of flags temporarily - I do plan to fix the codebase at a later point and toggle this on, but let's go step by step.

Here's the list with the reason:
- undocumented public documentation: it correctly realizes that for `///` the documentation is undefined. not sure yet whether this is a bug. Should we write `/// ditto`?
- "Local imports should specify the symbols being imported to avoid hiding local symbols": there were too many violations.
- unused variable: issue [#326](https://github.com/Hackerpilot/Dscanner/issues/326)
- could be immutable: issue [#325](https://github.com/Hackerpilot/Dscanner/issues/325)
- checks for labels with the same name as variables: issue [#323](https://github.com/Hackerpilot/Dscanner/issues/323)

source/mir/combinatorics/package.d
```diff
@@ -167,6 +166,24 @@ unittest
...
+*/
+IndexedRoR!(Collection, Range) indexedRoR(Collection, Range)(Collection c, Range r)
+if (isInputRange!Range)
+{
+    return IndexedRoR!(Collection, Range)(c, r);
+}
+
+/// ditto
 struct IndexedRoR(Collection, Range)
 if (isInputRange!Range)
```
wilzbach: this didn't have docs ...

source/mir/combinatorics/package.d
```diff
@@ -308,42 +274,13 @@ Permutations makePermutations(Allocator)(auto ref Allocator alloc, size_t n)
...
-    static immutable expected2 = [[0, 1], [1, 0]];
-    auto r = iota(2);
-    auto rp = makePermutations(Mallocator.instance, r.length);
-    assert(rp.indexedRoR(r).equal!equal(expected2));
-    dispose(Mallocator.instance, rp);
-}
-
+/// ditto
 struct Permutations
 {
```
wilzbach, line 359: docs were missing for all the structs - the order is now

`method`
`makeMethod` /// ditto
struct `Method`  /// ditto
unittest
dispose(Method ..)

source/mir/las/sum.d
```diff
@@ -374,9 +374,13 @@ struct Summator(T, Summation summation)
...
 
     static if(summation == Summation.pairwise)
-        enum bool fastPairwise = 
-            isFloatingPoint!F || 
-            isComplex!F || 
+        /**
+        Fast pairwise is enabled for floating point, complex numbers.
+        If SimD is enabled it also works for other numeric types.
+        */
+        enum bool fastPairwise =
```
wilzbach: public method - requires doc. can u check whether this makes sense?

source/mir/las/sum.d
```diff
@@ -374,9 +374,13 @@ struct Summator(T, Summation summation)
...
+            isFloatingPoint!F ||
+            isComplex!F ||
```
wilzbach, line 379: trailing whitespace ...

source/mir/las/sum.d
```diff
@@ -1822,16 +1828,20 @@ nothrow unittest
...
-            assert(summator.isFinite() == r.isFinite() || r == -double.max && s == -double.infinity || r == double.max && s == double.infinity);
-            assert(summator.isInfinity() == r.isInfinity() || r == -double.max && s == -double.infinity || r == double.max && s == double.infinity);
-            assert(nextDown(s) <= r && r <= nextUp(s) || s.isNaN && r.isNaN);
+            assert(summator.isFinite() == r.isFinite()
+                 || (r == -double.max && s == -double.infinity)
+                 || (r == double.max && s == double.infinity));
+            assert(summator.isInfinity() == r.isInfinity()
+                 || (r == -double.max && s == -double.infinity)
+                 || (r == double.max && s == double.infinity));
+            assert((nextDown(s) <= r && r <= nextUp(s)) || (s.isNaN && r.isNaN));
```
wilzbach: dscanner checks for such tricky cases where the prevalence of `&&` over `||` matters.

source/mir/las/sum.d
```diff
@@ -2204,7 +2214,9 @@ private template SummationType(F)
     version(X86) //workaround for Issue 13474
     {
         static if (!is(Unqual!F == real) && (isComplex!F || !is(Unqual!(typeof(F.init.re)) == real)))
-            pragma(msg, "Note: Summation algorithms on x86 use 80bit representation for single and double floating point numbers.");
+            enum note =  "Note: Summation algorithms on x86 use 80bit representation"
+                       ~ "for single and double floating point numbers.";
+            pragma(msg, note);
```
wilzbach, line 2212: over 120 chars

source/mir/ndslice/iteration.d
```diff
@@ -127,12 +127,10 @@ template swapped(size_t dimensionA, size_t dimensionB)
     auto swapped(size_t N, Range)(Slice!(N, Range) slice)
     {
         {
-            enum i = 0;
```
wilzbach, line 130: unused variable ...

9il:
> Naming style guide

turn off please

source/mir/ndslice/selection.d
```diff
@@ -917,7 +917,7 @@ auto byElement(size_t N, Range)(auto ref Slice!(N, Range) slice)
             size_t[N] _indexes;
 
             static if (canSave!PureRange)
-            auto save() @property
+            auto save()
```
wilzbach: `@property` methods with zero arguments should return `const`.
Changing this as it still can be used via UCFS

source/mir/ndslice/selection.d
```diff
@@ -1686,12 +1686,12 @@ template IotaSlice(size_t N)
 
 // undocumented
 // zero cost variant of `std.range.iota`
-struct IotaMap()
+private struct IotaMap()
```
wilzbach: public method without documentation should be set to private or protected

9il:
I would not merge this until the Phobos merged or you remove changes in `ndslice` package

9il:
Please move all changes to separate PR (and without ndslice)

source/mir/las/sum.d
9il: should be private

wilzbach:
@9il I tried to describe why I did some of the changes - hope this helps to understand ;-)
Will separate now.

wilzbach:
> Please move all changes to separate PR (and without ndslice)

I rebased this one and moved the ndslice changes to #89 which you can keep open.

9il:
split any changes from linter addition

wilzbach:
> split any changes from linter addition

done.

source/mir/combinatorics/package.d
```diff
@@ -265,39 +264,6 @@ if (isInputRange!Range && hasLength!Range)
...
-    // is copyable?
-    import std.algorithm: equal;
-    import std.range: iota, dropOne;
-    auto a = iota(2).permutations;
-    assert(a.front.equal([0, 1]));
-    assert(a.save.dropOne.front.equal([1, 0]));
-    assert(a.front.equal([0, 1]));
-}
-
 /// ditto
```
9il, line 335: should have separate comment

source/mir/combinatorics/package.d
```diff
@@ -308,42 +274,13 @@ Permutations makePermutations(Allocator)(auto ref Allocator alloc, size_t n)
...
-
-    static immutable expected2 = [[0, 1], [1, 0]];
-    auto r = iota(2);
-    auto rp = makePermutations(Mallocator.instance, r.length);
-    assert(rp.indexedRoR(r).equal!equal(expected2));
-    dispose(Mallocator.instance, rp);
-}
-
+/// ditto
 struct Permutations
```
9il, line 358: should have separate comment

source/mir/combinatorics/package.d
```diff
@@ -308,42 +274,13 @@ Permutations makePermutations(Allocator)(auto ref Allocator alloc, size_t n)
...
 {
     private uint[] indices, state;
     private bool _empty;
 
-    this(uint[] state, uint[] indices) @safe pure nothrow @nogc
+    private this(uint[] state, uint[] indices) @safe pure nothrow @nogc
```
9il: make it public please with `///`

source/mir/las/sum.d
9il: turn it off please

source/mir/combinatorics/package.d
wilzbach, line 358: "Forward range of Permutations?"

Or more detailed? Most of the stuff is already explained at the `permutations` method, hence the `ditto`.

source/mir/combinatorics/package.d
wilzbach: But then we should explain that those arrays just need to be empty?

9il: no, just need to have proper size

source/mir/combinatorics/package.d
9il, line 358: > Or more detailed

A little bit more and links to make**\* and ***

9il, line 358: See_also: links

wilzbach, line 358: already in the working - how can i link to the according `dispose`?

I know only `$(LREF dispose)`

9il:
BTW, i think documentation is more important now because we included combinatorics and will include sparse soon.

9il:
only ndslice has docs on dlang.org

source/mir/combinatorics/package.d
wilzbach, line 335: was this note for `Permutations` - imho `makePermutations` should be bundled with `permutations`

source/mir/combinatorics/package.d
```diff
@@ -410,6 +366,70 @@ struct Permutations
...
+    auto rp = permutations(r.length).indexedRoR(r);
+    assert(rp.equal!equal(expectedRes));
+
+    // direct style
+    auto rp2 = iota(3).permutations;
+    assert(rp2.equal!equal(expectedRes));
+}
+
+///
+static if (__VERSION__ > 2069) @nogc unittest
```
wilzbach, line 472: I think as long as allocators are still new we should make the unittest public.

source/mir/combinatorics/package.d
```diff
@@ -308,41 +277,28 @@ Permutations makePermutations(Allocator)(auto ref Allocator alloc, size_t n)
...
-{
-    import std.algorithm: equal;
-    import std.range : iota;
+/**
+Lazy Forward range of permutations using $(WEB
+en.wikipedia.org/wiki/Heap%27s_algorithm, Heap's algorithm).
 
-    import std.experimental.allocator.mallocator;
+It will always generate the permutations from 0 to `n - 1`,
+use $(LREF indexedRoR) to map it to your range.
```
wilzbach, line 350: how do you like this sentence - is it clear?

9il, line 350: my english is not well, but it seems that present simple is better then future

wilzbach, line 350: > my english is not well,

no worries - your coding is well!
Removed the `will`

source/mir/combinatorics/package.d
```diff
@@ -308,41 +277,28 @@ Permutations makePermutations(Allocator)(auto ref Allocator alloc, size_t n)
...
-static if (__VERSION__ > 2069) @nogc unittest
-{
-    import std.algorithm: equal;
-    import std.range : iota;
+/**
+Lazy Forward range of permutations using $(WEB
+en.wikipedia.org/wiki/Heap%27s_algorithm, Heap's algorithm).
 
-    import std.experimental.allocator.mallocator;
+It always generate the permutations from 0 to `n - 1`,
```
9il: generateS?

wilzbach: refresh :)

source/mir/combinatorics/package.d
9il, line 335: hmm. ok

source/mir/las/sum.d
wilzbach: Are u sure? it does make the code better to read. The idea is to avoid confusing expressions:

```
if (a && b || c) // bad
if (a && (b || c)) // good
```

It checks is if there are multiple `||` and `&&` in one expression,
so `1 && 2 && 3` will not trigger it.

It is called ["logical precedence check"](https://github.com/Hackerpilot/Dscanner/blob/master/src/analysis/logic_precedence.d)

9il: no, i don't like braces, expressions become less readable

9il: > Are u sure?

Yes

wilzbach: > no, i don't like braces, expressions become less readable

I disagree, but you are the "boss" ;-)
Removed.

wilzbach:
> BTW, i think documentation is more important now because we included combinatorics and will include sparse soon.

We will have a look during the weekend ;-)

wilzbach:
It seems like I "force-pushed" travis to its limits, but now it's passing finally :)

source/mir/combinatorics/package.d
```diff
@@ -167,6 +166,24 @@ unittest
     assert(binomial!BigInt(-5, -7) == 15);
 }
 
+/**
+Creates a projection of `collection` Range (e.g. `Combinations`)
+onto a custom `range`.
+
+Params:
+    collenction = Range to be projected from
```
9il: should be lower case

source/mir/combinatorics/package.d
```diff
@@ -308,41 +277,28 @@ Permutations makePermutations(Allocator)(auto ref Allocator alloc, size_t n)
...
-    auto r = iota(2);
-    auto rp = makePermutations(Mallocator.instance, r.length);
-    assert(rp.indexedRoR(r).equal!equal(expected2));
-    dispose(Mallocator.instance, rp);
-}
+Generating a new item is in `O(k)` (amortized `O(1)`),
+the total number of elements is `n^k`.
 
+See_Also:
+    $(LREF cartesianPower), $(LREF makeCartesianPower)
```
9il: Copy past error

source/mir/combinatorics/package.d
```diff
@@ -251,6 +250,9 @@ Params:
 
 Returns:
     Forward range, which yields the permutations
+
+See_Also:
+    $(LREF dispose)
```
9il: remove $(LREF dispose) and add $(LREF Permutations)

source/mir/combinatorics/package.d
```diff
@@ -251,6 +300,9 @@ Params:
 
 Returns:
     Forward range, which yields the permutations
+
+See_Also:
+    $(LREF Permutations), $(LREF dispose)
```
wilzbach: As it is bundled with `makePermutations`, linking to `dispose` could be useful to the user

source/mir/combinatorics/package.d
```diff
@@ -265,41 +317,8 @@ if (isInputRange!Range && hasLength!Range)
...
-    import std.range: iota, dropOne;
-    auto a = iota(2).permutations;
-    assert(a.front.equal([0, 1]));
-    assert(a.save.dropOne.front.equal([1, 0]));
-    assert(a.front.equal([0, 1]));
-}
-
 /// ditto
-Permutations makePermutations(Allocator)(auto ref Allocator alloc, size_t n)
+Permutations makePermutations(Allocator)(size_t n, auto ref Allocator alloc)
```
wilzbach: had the wrong order ;-)

source/mir/combinatorics/package.d
9il: `dispose` make refer to another `dispose`

source/mir/combinatorics/package.d
```diff
@@ -359,6 +366,10 @@ struct Permutations
         this.state = state;
 
         _empty = indices.length == 0;
+
+        // factorial
+        foreach (i; 1..indices.length + 1)
+            _max_states *= i;
```
wilzbach, line 387: all the other ranges provide a `length` function - i don't think it makes a noticeable overhead. Alternatively we could also compute it on-demand only

source/mir/combinatorics/package.d
9il: No, allocator should be always first for make\* functions

source/mir/combinatorics/package.d
wilzbach: > dispose make refer to another dispose

Ehm sry - what do you mean?

source/mir/combinatorics/package.d
wilzbach: Ok will change the other functions

wilzbach:
Hmm it seems that there were still a lot of stuff to do for the combinatorics module - here's a list of what i changed:
- allocator order for makeCombinations,Repeat,CartesianPower
- length for permutations (all other collections have this too)
- unittests for length
- unittest for invalid input
- unittest for indexed
- added const to `@property` attributes like `bool` and `length`
- reworked the description for `IndexedRoR

9il:
> allocator order for makeCombinations,Repeat,CartesianPower

allocators should be first argumensts

source/mir/combinatorics/package.d
9il: you have multiple dispose declarations

wilzbach:
> allocators should be first argumensts

Yep those had the allocater as last argument

source/mir/combinatorics/package.d
wilzbach: So there is no way to link to the "correct" one?

9il: no way

wilzbach: I just tested and if I click on `dispose` it will get to such a page - is this good enough or should we trust that the unittest does the explanation?

image

9il: this is not good enough

source/mir/combinatorics/package.d
wilzbach: already fixed - note at the wrong place?

source/mir/combinatorics/package.d
```diff
@@ -613,14 +718,59 @@ pure nothrow @safe unittest
...
+
+/**
+Lazily computes all k-combinations of `r`.
 Imagine this as the $(LREF cartesianPower) filtered for only strictly ordered items.
 
 While generating a new combination is in `O(k)`,
 the number of combinations is `binomial(n, k)`.
 
-For example $(D"AB".combinations(2).array) returns $(D["AB"]).
+For example `"AB".combinations(2).array` returns `["AB"]`.
```
wilzbach: should I remove `.array` here?

wilzbach:
@9il CI was failing due to an added tests which included the `Mallocator` - should be fixed now. Any other objections? :)

source/mir/combinatorics/package.d
9il: remove this example from the docs. And add constraints, that templates are note accepts narrow strings.

wilzbach:
> remove this example from the docs. And add constraints, that templates are note accepts narrow strings.

I created a module level unittest as an overview over the iterators.
Moreover I realized that `indexed` from std.range actually needs a random access range and thus corrected all template constraints. A narrow string can't be a random access string :)

wilzbach:
(rebased to latest master - as we are moving quite fast tonight)

wilzbach:
> Failed to download http://code.dlang.org/packages/libevent/2.0.1+2.0.16.zip: 404 Not Found

why? :(

9il:
> why? :(

Just remove ddox. I don't like vibe.d and ddox ;/

We can use https://github.com/kiith-sa/harbored-mod, it is the best but needs small update. Probably we would need to add macros file from dlang

wilzbach:
> We can use https://github.com/kiith-sa/harbored-mod, it is the best but needs small update.

Okay will send a PR later ;-)

wilzbach:
> Okay will send a PR later ;-)

Hmm seems that we have to fix libdparse first ...
https://github.com/kiith-sa/harbored-mod/issues/62

PR #89, opened 2016-04-14
fix style syntax for mir.ndslice

wilzbach:
Pending changes for ndslice from #88

### DIFF
--- a/source/mir/ndslice/iteration.d
+++ b/source/mir/ndslice/iteration.d
@@ -250,7 +250,8 @@ template rotated(size_t dimensionA, size_t dimensionB)
 }
 
 /// ditto
-Slice!(N, Range) rotated(size_t N, Range)(Slice!(N, Range) slice, size_t dimensionA, size_t dimensionB, sizediff_t k = 1)
+Slice!(N, Range) rotated(size_t N, Range)(Slice!(N, Range) slice,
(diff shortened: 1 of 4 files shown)


source/mir/ndslice/iteration.d
```diff
@@ -127,12 +127,10 @@ template swapped(size_t dimensionA, size_t dimensionB)
     auto swapped(size_t N, Range)(Slice!(N, Range) slice)
     {
         {
-            enum i = 0;
             alias dimension = dimensionA;
             mixin DimensionCTError;
         }
         {
-            enum i = 1;
```
9il, line 135: BTW, have you looked at `DimensionCTError`?

wilzbach, line 135: damn - i just looked at the first one. My mistake :/

```
    static assert(dimension < N,
        "dimension = " ~ dimension.stringof ~ " at position "
        ~ i.stringof ~ " should be less than N = " ~ N.stringof
        ~ tailErrorMessage!());
```

9il:
i can not merge this until phobos PR, but i can merge fixes for sum and combinatorics

wilzbach:
I understood, #88 only contains those. Dscanner will only print a couple of warnings which we can resolve as soon as ndslice is open for changes.

wilzbach:
This is exactly similar to #154 - closing in favor of the newer PR.

Issue #49, opened 2016-04-10
mir.stat: statistical functions 


wilzbach:
### TODO
- Probabilistic counting. See https://en.wikipedia.org/wiki/HyperLogLog

Statistical functions and algorithms. 

NumPy's capabilities:
https://github.com/DlangScience/mir/wiki/NumPy:-Statistics
##

9il:
> Maybe we can reuse parts from dstats?

We can reuse my `atmosphere`, but it is small. `dstats` has not BSL license :-/

wilzbach:
> Each file contains a license header. All modules that are exclusively written by the main author (David Simcha) are licensed under the Boost license, so that pieces of them may freely be incorporated into Phobos 

(From dstats) 

I haven't checked details though.

9il:
OK, that is good :-)

9il:
no plan to add the package into the main Mir repo

BTW,  https://github.com/tamediadigital/hll-d

PR #90, opened 2016-04-14
mir.stat: statistical functions

wilzbach:
Yesterday I realized that `mir.random` heavily depends on `mir.stat` as it needs statistical tests to verify its random algorithsm.
Well here is the WIP part for this - it includes already includes an online variance + mean report, moreover I plan to add one or two statistical tests in the first version. What else do you want to see in the initial version of `mir.stat` (#49)?

TODO:
- more precision tests

### DIFF
--- a/source/mir/stat/package.d
+++ b/source/mir/stat/package.d
@@ -1,0 +1,209 @@
+/**
+Statistical functions
+
+License: $(LINK2 http://boost.org/LICENSE_1_0.txt, Boost License 1.0).
+*/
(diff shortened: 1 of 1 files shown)


9il:
Don't spend time on this.

The better option for testing RNGs is a collection of PDF functions, drawing tool, and log-likelihood. Plus articles!!!

9il:
And do not work on best API. It would be reworked, maybe few times

PR #99, opened 2016-04-15
fix #97 - failing uint on windows x86

### LINKED ISSUE #97: mir.combinatorics: uint on windows x86
seems like I should add some casts - will prepare a PR.

```
source\mir\combinatorics\package.d(129,27): Error: safe function 'mir.combinatorics.__unittestL122_2' cannot call system function 'mir.combinatorics.binomial!(BigInt, int).binomial' 
source\mir\combinatorics\package.d(700,35): Error: cannot implicitly convert expression (binomial(n, repeatLen)) of type ulong to uint
source\mir\combinatorics\package.d(983,34): Error: cannot implicitly convert expression (binomial(n + repeatLen - 1u, repeatLen)) of type ulong to uint
```

wilzbach:
I can't reproduce or understand this error - do you have an idea?

```
source\mir\combinatorics\package.d(129,27): Error: safe function 'mir.combinatorics.__unittestL122_2' cannot call system function 'mir.combinatorics.binomial!(BigInt, int).binomial' 
```

PR fixes the other two errors in #97, even though I think that we should use `ulong` for the `max_states` by default ...

### DIFF
--- a/source/mir/combinatorics/package.d
+++ b/source/mir/combinatorics/package.d
@@ -119,13 +119,13 @@ R binomial(R = ulong, T)(T n, T k)
 }
 
 ///
-pure @safe unittest
+pure unittest
 {
     assert(binomial(5, 2) == 10);
     assert(binomial(6, 4) == 15);
     assert(binomial(3, 1) == 3);
 
-    import std.bigint;
+    import std.bigint: BigInt;
     assert(binomial!BigInt(1000, 10) == BigInt("263409560461970212832400"));
 }
 
@@ -697,7 +697,7 @@ public:
 
         uint repeatLen = cast(uint) state.length;
         this.n = cast(uint) n;
-        this.max_states = binomial(n, repeatLen);
+        this.max_states = cast(size_t) binomial(n, repeatLen);
         this.state = state;
 
         // set initial state and calculate max possibilities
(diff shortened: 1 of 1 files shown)


source/mir/combinatorics/package.d
```diff
@@ -697,7 +697,7 @@ public:
 
         uint repeatLen = cast(uint) state.length;
         this.n = cast(uint) n;
-        this.max_states = binomial(n, repeatLen);
+        this.max_states = cast(uint) binomial(n, repeatLen);
```
9il: cast(size_t)

source/mir/combinatorics/package.d
```diff
@@ -980,7 +980,7 @@ public:
         // set initial state and calculate max possibilities
         if (n > 0)
         {
-            max_states = binomial(n + repeatLen - 1, repeatLen);
+            max_states = cast(uint) binomial(n + repeatLen - 1, repeatLen);
```
9il: ditto

9il:
> I can't reproduce or understand this error - do you have an idea?

Yes, on 32 bit systems bigint uses assembler version. So just remove `@safe` for unittest

wilzbach:
> cast(size_t)

arg. Stupid mistake. sorry!

> Yes, on 32 bit systems bigint uses assembler version. So just remove @safe for unittest

Wow. Thanks!

(rebased with master)

PR #100, opened 2016-04-15
fix #98 - use -m64 for 64 bit windows

### LINKED ISSUE #98: appveyor.yml should contain -m64 flag for 64 bit windows

wilzbach:
This should in theory fix #98 - I need to wait a couple of minutes until AppVeyor builds, so will let it open ;-)

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -22,7 +22,8 @@ install:
   - dub --version
 
 build_script:
- - echo build
+ - echo dummy build script - dont remove me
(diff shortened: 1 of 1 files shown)


9il:
BTW, why `unittest-cov`??

wilzbach:
> BTW, why unittest-cov??

I originally planned to upload the result to codecov too, but that seems a bit more difficult. Changed ;-)

Issue #101, opened 2016-04-15
https://circleci.com configuration file


9il:

9il:
it has not D support :-/

wilzbach:
> it has not D support :-/

we can download it - see #102 ;-)

PR #102, opened 2016-04-15
Circleci integration

wilzbach:
CircleCI integration (#101).
It seems to be pretty fast (<30s)!

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -1,0 +1,15 @@
+machine:
+  environment:
+    DMD: 2.071.0
+    DUB: 0.9.24
+    PATH: "${HOME}/dmd2/linux/bin64:${PATH}"
+    LD_LIBRARY_PATH: "${HOME}/dmd2/linux/lib64:${LD_LIBRARY_PATH}"
+dependencies:
+  override:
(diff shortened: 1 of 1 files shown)


wilzbach:
See https://circleci.com/gh/wilzbach/mir

When you enable it via the web interface - there is also a setting to allow up to 4 parallel builds :)

circle.yml
```diff
@@ -0,0 +1,15 @@
+machine:
+  environment:
+    DMD: 2.071.0
+    DUB: 0.9.24
+    PATH: "${HOME}/dmd2/linux/bin64:${PATH}"
+    LD_LIBRARY_PATH: "${HOME}/dmd2/linux/lib64:${LD_LIBRARY_PATH}"
```
9il, line 6: OS X support?

wilzbach, line 6: Not possible in the free version - https://circleci.com/pricing/#build-linux

9il, line 6: oh, ok

wilzbach:
Btw at some point we can also setup a [build matrix](https://github.com/michaelcontento/circleci-matrix) there, but it seems to be more complicated and for now we can enjoy the fast success / error from within one minute :)

Issue #103, opened 2016-04-16
fix circle doc setup


wilzbach:
i guess the easiest solution for now is just to host a binary of ddox?

9il:
No, just remove ddox build

wilzbach:
> No, just remove ddox build

Ok, but only until we found a better doc engine :/

9il:
> Ok, but only until we found a better doc engine :/

I will fix harbored

9il:
> I will fix harbored

I have fixed bugs, but it has a lot of visual issues. So we need normal DDOC with macros from dlang.org

wilzbach:
https://github.com/dlang/dlang.org/blob/master/html.ddoc

wilzbach:
And https://github.com/dlang/dlang.org/blob/master/std.ddoc +
https://github.com/dlang/dlang.org/blob/97d3e5eff143c655f70008066145568b91e2a3d6/std-ddox.ddoc

Seems like a lot of files  :(

9il:
We should go with bootDoc thought: https://github.com/JakobOvrum/bootDoc

9il:
See also http://jakobovrum.github.io/bootdoc-phobos/ =) This already contains macros!!!

wilzbach:
But this runs ddox in the background and just adds a couple of css classes?

9il:
No the DDOX, this is common DDOC (by DMD)

wilzbach:
> No the DDOX, this is common DDOC (by DMD)

Ah I see  - sorry about the confusion :/

9il:
try `$ rdmd bootDoc/generate.d --output=docs source` from master

9il:
this should generate docs in folder `docs`.

wilzbach:
hmm - did you see that?
![image](https://cloud.githubusercontent.com/assets/4370550/14580990/4b88cd66-03e9-11e6-8acb-a70bdaa551e7.png)

wilzbach:
Apart from that I really like it :)
Btw is there a way to generate a overview or main page? 
Would be nice to have, so that I can setup the s3 integration again.

9il:
> hmm - did you see that?

What wrong with that?

> Btw is there a way to generate a overview or main page? 
> Would be nice to have, so that I can setup the s3 integration again.

We should generate it from a marcdown file, currently main page for reference is `mir.html`, you can change it in `settings.ddoc`

[...]

PR #104, opened 2016-04-16
remove ddox setup

wilzbach:
Ideally this should already solve our problem at #103


wilzbach:
Removed entire ddox setup - will be restored once we find a better solution ...

Issue #105, opened 2016-04-16
notice about extensive testing in ReadMe


wilzbach:
Probably we want to add a small message like that once Travis and AppVeyor are fixed:

> Mir is automatically tested with recent releases of dmd and ldc for Linux, OS X and Windows x86 and x64.

9il:
We also need to test 32 bit linux

wilzbach:
I think nownow we can put the notice in the wiki? Or is the fact that ldc under Windows doesn't compile a blocker?

9il:
Yes, we can, but with note about LDC issues. Some matrix/table would be helpful for each compiler

wilzbach:
added with #153

PR #106, opened 2016-04-16
test 32bit on travis

wilzbach:
Apparently it is as easy as adding `--arch x86`.

Please note that the [build matrix](https://docs.travis-ci.com/user/customizing-the-build/#Build-Matrix) now expands quite heavily, see [here](https://travis-ci.org/wilzbach/mir/builds/123489045)

Imho that's ok, because we have circleci as a fast check.

Also mir.las.sum seems to fail, see https://travis-ci.org/wilzbach/mir/jobs/123489047

(related to #105)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,14 +10,22 @@ d:
  - dmd-2.068.2
  - ldc-1.0.0-alpha1
  - ldc-0.17.0
-install:
- - if [[ $(dmd --help | head -n 1 | cut -d ' ' -f 4) == 'v2.071.0' && $TRAVIS_OS_NAME == 'linux' ]] ; then export IS_LATEST_PHOBOS=1 ; fi
- - wget -O doveralls "https://github.com/ColdenCullen/doveralls/releases/download/v1.2.0/doveralls_linux_travis"
- - chmod +x doveralls
- - wget -O dscanner "http://releases.dlang.io/dscanner/0.3.0-git/dscanner"
- - chmod +x dscanner
+env:
+ - ARCH="x86_64"
+matrix:
+  include:
+    - {os: linux, d: dmd-2.071.0, env: ARCH="x86", addons: {apt: {packages: [[gcc-multilib]]}}}
+    - {os: linux, d: dmd-2.070.2, env: ARCH="x86", addons: {apt: {packages: [[gcc-multilib]]}}}
+    - {os: linux, d: dmd-2.069.2, env: ARCH="x86", addons: {apt: {packages: [[gcc-multilib]]}}}
+    - {os: linux, d: dmd-2.068.2, env: ARCH="x86", addons: {apt: {packages: [[gcc-multilib]]}}}
+    - {os: linux, d: ldc-1.0.0-alpha1, env: ARCH="x86", addons: {apt: {packages: [[gcc-multilib]]}}}
+    - {os: linux, d: ldc-0.17.0, env: ARCH="x86", addons: {apt: {packages: [[gcc-multilib]]}}}
+branches:
+  only:
+    - master
+    - stable
 script:
- - if [ "$IS_LATEST_PHOBOS" ] ; then ./dscanner --config .dscanner.ini --styleCheck source 2> /dev/null ; else echo "NOT_LATEST"; fi
- - dub test -b unittest-cov
+ - echo "$ARCH"
+ - dub test --arch "$ARCH" -b unittest-cov
 after_success:
  - bash <(curl -s https://codecov.io/bash)


9il:
1. no need to 32 bit tests for osx
2. no need `-b unittest-cov` for all targets except one for codecov

wilzbach:
Why did you close this?

> no need to 32 bit tests for osx

why?

> no need -b unittest-cov for all targets except one for codecov

done.

9il:
> Why did you close this?

Sorry, i need to sleep)))

> why?

OS X is 64 bit only. 32 bit target is deprecated

.travis.yml
```diff
@@ -10,14 +10,26 @@ d:
...
+ - x86_64_BUILD=TRUE
+ - x86_32_BUILD=TRUE
+matrix:
+  exclude:
+    - osx
+    - x86_32_BUILD=TRUE
+addons:
+  apt:
+    packages:
+        gcc-multilib
```
9il: only for 32 `x86_32_BUILD`

wilzbach: yep I know - I am still trying to teach travis that ...

9il: the same way somehow should work for `script`. i hope so

wilzbach:
Wohoo - I think I got it now working for all Linux 32bit targets :)

At least it compiles now, however unittests are still failing :/

https://travis-ci.org/wilzbach/mir/jobs/123495990#L306

.travis.yml
wilzbach: I like the way it is done now - it adds only a minimal overhead and we can exactly pick which versions we want to test under 32bit.

wilzbach:
Update: #109 is the only remaining failing test - maybe you reduce the tested precision for x86?

.travis.yml
```diff
@@ -10,14 +10,26 @@ d:
...
+    - {os: linux, d: dmd-2.068.2, env: ARCH="x86", addons: {apt: {packages: [[gcc-multilib]]}}}
+    - {os: linux, d: ldc-1.0.0-alpha1, env: ARCH="x86", addons: {apt: {packages: [[gcc-multilib]]}}}
+    - {os: linux, d: ldc-0.17.0, env: ARCH="x86", addons: {apt: {packages: [[gcc-multilib]]}}}
 install:
  - if [[ $(dmd --help | head -n 1 | cut -d ' ' -f 4) == 'v2.071.0' && $TRAVIS_OS_NAME == 'linux' ]] ; then export IS_LATEST_PHOBOS=1 ; fi
  - wget -O doveralls "https://github.com/ColdenCullen/doveralls/releases/download/v1.2.0/doveralls_linux_travis"
  - chmod +x doveralls
  - wget -O dscanner "http://releases.dlang.io/dscanner/0.3.0-git/dscanner"
  - chmod +x dscanner
+ - if [ "$x86_32_BUILD" ] ; then apt-get install gcc-multilib ; fi
```
9il: 1. should be done with addons
2. only for linux

.travis.yml
```diff
@@ -10,14 +10,26 @@ d:
...
 script:
  - if [ "$IS_LATEST_PHOBOS" ] ; then ./dscanner --config .dscanner.ini --styleCheck source 2> /dev/null ; else echo "NOT_LATEST"; fi
- - dub test -b unittest-cov
+ - echo "$ARCH"
+ - if [ "$IS_LATEST_PHOBOS" ] && [ "$ARCH" == "x86_64" ] ; then dub test -b unittest-cov ; else dub test --arch "$ARCH" ; fi
```
9il: Please avoid all ifs. It is evil

wilzbach: You wrote yesterday night that it should only be executed once...

9il: yes, should be one.
1. This is executed twice (fox Limux and OS X)
2. This is very buggy

9il: we need to find another way for matrix defenotion

wilzbach: hmm I actually thought that it is good that we upload both coverage reports to coveralls - as we might have different ones for 32bit

.travis.yml
wilzbach: I forgot to remove that.

wilzbach:
Idea: we do all the ifs (coverage and dscanner) with circleci? Then we don't have any ugliness in the travis.yml and circleci is pretty fast anyways. You just should set this then to enforced too. I will prepare another PR ;-)

9il:
ok

.travis.yml
```diff
@@ -10,14 +10,16 @@ d:
...
  - ldc-1.0.0-alpha1
  - ldc-0.17.0
-install:
- - if [[ $(dmd --help | head -n 1 | cut -d ' ' -f 4) == 'v2.071.0' && $TRAVIS_OS_NAME == 'linux' ]] ; then export IS_LATEST_PHOBOS=1 ; fi
- - wget -O doveralls "https://github.com/ColdenCullen/doveralls/releases/download/v1.2.0/doveralls_linux_travis"
- - chmod +x doveralls
- - wget -O dscanner "http://releases.dlang.io/dscanner/0.3.0-git/dscanner"
- - chmod +x dscanner
+env:
+ - ARCH="x86_64"
```
9il, line 14: do not forget to define 2 archs

wilzbach, line 14: see below at the matrix includes - `x86_64` is our default architecture and we include a couple of 32bit ones ;-)

.travis.yml
```diff
@@ -10,14 +10,20 @@ d:
...
+    - {os: linux, d: dmd-2.071.0, env: ARCH="x86", addons: {apt: {packages: [[gcc-multilib]]}}}
+    - {os: linux, d: dmd-2.070.2, env: ARCH="x86", addons: {apt: {packages: [[gcc-multilib]]}}}
+    - {os: linux, d: dmd-2.069.2, env: ARCH="x86", addons: {apt: {packages: [[gcc-multilib]]}}}
+    - {os: linux, d: dmd-2.068.2, env: ARCH="x86", addons: {apt: {packages: [[gcc-multilib]]}}}
+    - {os: linux, d: ldc-1.0.0-alpha1, env: ARCH="x86", addons: {apt: {packages: [[gcc-multilib]]}}}
+    - {os: linux, d: ldc-0.17.0, env: ARCH="x86", addons: {apt: {packages: [[gcc-multilib]]}}}
+branches:
+  only:
+    - master
+    - stable
```
wilzbach, line 26: this avoids that PR are built twice ...

wilzbach:
Rebased to master - ideally [travis](https://travis-ci.org/DlangScience/mir/builds/123529930) should pass now :)

.travis.yml
```diff
@@ -10,14 +10,20 @@ d:
...
+env:
+ - ARCH="x86_64"
+matrix:
+  include:
+    - {os: linux, d: dmd-2.071.0, env: ARCH="x86", addons: {apt: {packages: [[gcc-multilib]]}}}
+    - {os: linux, d: dmd-2.070.2, env: ARCH="x86", addons: {apt: {packages: [[gcc-multilib]]}}}
+    - {os: linux, d: dmd-2.069.2, env: ARCH="x86", addons: {apt: {packages: [[gcc-multilib]]}}}
+    - {os: linux, d: dmd-2.068.2, env: ARCH="x86", addons: {apt: {packages: [[gcc-multilib]]}}}
+    - {os: linux, d: ldc-1.0.0-alpha1, env: ARCH="x86", addons: {apt: {packages: [[gcc-multilib]]}}}
+    - {os: linux, d: ldc-0.17.0, env: ARCH="x86", addons: {apt: {packages: [[gcc-multilib]]}}}
```
9il, line 22: Can this be done more elegant way?

wilzbach, line 22: Well we can just run the 32bit on the 64bit Arch - there's actually no point in spinning up another container for this. If one clicks at the log, he would still see what test is failing and if we don't use so many builds travis will be faster again - How do you like this idea?

9il, line 22: it is more clear to have separate block

wilzbach:
Yeah we are [passing for 32bit](https://travis-ci.org/DlangScience/mir/builds/123529930) - now I just need to put codecov back.

.travis.yml
wilzbach, line 22: so we stay with this config?

wilzbach:
> now I just need to put codecov back.

I tried to avoid the `if`, so we upload the coverage report on all architectures to coveralls (like before).

Btw for what do we need `doveralls`?

wilzbach:
> Btw for what do we need doveralls?

Update: we don't need it - removed :)

9il:
> Update: we don't need it - removed :)

Why, it uploads coverage to codecov

wilzbach:
> Why, it uploads coverage to codecov

Nope `coveralls` supports `D`out of the box - see [this current log](https://travis-ci.org/DlangScience/mir/jobs/123531550)

wilzbach:
Alrighty - Travis is all green :)

PR #113, opened 2016-04-16
Fix107

9il:


wilzbach:
Wow that's elegant - just go for it. It works on my system too ;-)

PR #112, opened 2016-04-16
call codecov and dscanner with circleci

wilzbach:
In reference to #106 - I still need to verify that it actually works.

PR #114, opened 2016-04-16
only run dub test on circleci

wilzbach:
follow-up to #112

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -15,4 +15,4 @@ dependencies:
 test:
   override:
     - ./dscanner --config .dscanner.ini --styleCheck source 2> /dev/null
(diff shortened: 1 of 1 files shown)


wilzbach:
Btw don't forgot to set CircleCI to `required` too ;-)

9il:
> Btw don't forgot to set CircleCI to required too ;-)
>  View details

I turned of this for me XP

wilzbach:
> I turned of this for me XP

Is it possible to do this member-wise? 
So that I still have those constraints?

9il:
> Is it possible to do this member-wise? 
> So that I still have those constraints?

Yes, will turned off it for all

PR #118, opened 2016-04-16
Bootdoc

9il:


bootDoc
```diff
@@ -0,0 +1 @@
+Subproject commit f49a4177cfdccf793b649e7bf36d6eefa8583fa2
```
wilzbach, line 1: ?

.gitmodules
```diff
@@ -0,0 +1,3 @@
+[submodule "bootDoc"]
+	path = bootDoc
+	url = git://github.com/JakobOvrum/bootDoc.git
```
wilzbach, line 3: dub would be nicer, but I see it's not there :/

9il, line 3: No, this should be in the folder. So, this is the best option

PR #123, opened 2016-04-16
circleci s3 deployment

wilzbach:
this fixes our s3 setup again - I already set the AWS credentials [here](https://circleci.com/gh/DlangScience/mir/edit#aws) so we should be ready to go!

My test [worked](http://docs.mir.dlang.io/latest/mir.las.sum.html).

PS: It doesn't add the index html for the documentation, do you want to add a `index.html` to our setup?


wilzbach:
Thanks :)

PR #129, opened 2016-04-16
Update settings.ddoc

9il:


wilzbach:
I don't mind and this is probably the "Phobos" standard ;-)

PR #131, opened 2016-04-16
improve code style

wilzbach:
mostly 120L, public properties, nicer numbers, ...

PR #137, opened 2016-04-16
enable enforce option for codestyle

wilzbach:
after #131, we can "enforce" the code style.
As tonight you already commited a couple of violations (e.g. too long lines) and I did too, we should really have an annoying bot that reminds us.

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -15,7 +15,7 @@ dependencies:
     - git submodule update --recursive --init
 test:
   override:
-    - ./dscanner --config .dscanner.ini --styleCheck source 2> /dev/null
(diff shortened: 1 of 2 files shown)


circle.yml
```diff
@@ -15,7 +15,7 @@ dependencies:
     - git submodule update --recursive --init
 test:
   override:
-    - ./dscanner --config .dscanner.ini --styleCheck source 2> /dev/null
+    - ./dscanner --config .dscanner.ini --styleCheck --enforce $(find source/mir/* -type d | grep -v ndslice | tr '\n' ' ') 2> /dev/null
```
wilzbach: excludes the `ndslice` directory temporarily ;-)

wilzbach:
rebased, but let's wait for a minute whether circleci is passing ...

PR #139, opened 2016-04-16
Enable enforce codestyle

wilzbach:
Last time you merged to quickly - I was still experimenting with the settings. It turns out that there's a bug in Dscanner that always yields an exit code of 1.
I know this expression is a bit ugly now, but it works ...


wilzbach:
Thanks - from now on Circle CI will start to annoy you ;-)

PR #147, opened 2016-04-16
Update modules.ddoc

9il:


wilzbach:
Hmm - why do you want to hide the API as long as ndslice is still in Mir?

9il:
> Hmm - why do you want to hide the API as long as ndslice is still in Mir?
1. Documentation is much better in Phobos
2. Except sparse no other significant change will occur
3. We need to change macros in ndslice to support coorectly all links both in Mir and Phobos

wilzbach:
Okay I see your point with #148 - feel free to merge.

PR #150, opened 2016-04-17
use phobos style with dpldocs

wilzbach:
There are still a couple of bugs, but I guess one can see the direction.
I think it's mostly the links which don't work yet - [see it in action](http://docs.mir.dlang.io/phobos/test/).

__dont merge __ (Work in progress)

image

image

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,0 @@
-[submodule "bootDoc"]
-	path = bootDoc
-	url = git://github.com/DlangScience/bootDoc.git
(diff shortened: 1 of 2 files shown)


9il:
Why DDOX, but not DDOC? The last one is much more stable

wilzbach:
I just copied the setup from dlang.org..

9il:
If DDOX is good enough then do not spend time on DDOC

9il:
This is real shame that D has not good out of the box docs

wilzbach:
> If DDOX is good enough then do not spend time on DDOC

Well at least we can detect whether most macros are working correctly. It would be great to have a documentation as a service like many other languages do. 

I will continue to work on this once I find a bit of time again. My idea is now to be as close to the standard docs as possible for the ddoc macros.

wilzbach:
superseded by #203

PR #154, opened 2016-04-18
mir.ndslice: improve codestyle

wilzbach:
I still had these small improvement laying around - with them I can enable `dscanner` for `mir.ndslice` as it now passes our code style.

It's mostly just > 120 lines, missing `const`, better readability for numbers and undocumented methods.

### DIFF
--- a/source/mir/ndslice/selection.d
+++ b/source/mir/ndslice/selection.d
@@ -910,7 +910,7 @@ auto byElement(size_t N, Range)(auto ref Slice!(N, Range) slice)
             size_t[N] _indexes;
 
             static if (canSave!PureRange)
-            auto save() @property
+            auto save()
             {
                 return typeof(this)(_slice.save, _length, _indexes);
             }
@@ -1145,7 +1145,7 @@ auto byElement(size_t N, Range)(auto ref Slice!(N, Range) slice)
                 return typeof(return)(i, j);
             }
 
-            size_t[N] index() @property
+            size_t[N] index() @property const
             {
                 pragma(inline, true);
                 return _indexes;
@@ -1403,7 +1403,7 @@ auto byElementInStandardSimplex(size_t N, Range)(auto ref Slice!(N, Range) slice
             size_t[N] _indexes;
 
             static if (canSave!PureRange)
-            auto save() @property
+            auto save()
             {
                 return typeof(this)(_slice.save, _length, maxCobeLength, sum, _indexes);
             }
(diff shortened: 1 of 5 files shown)


source/mir/ndslice/selection.d
```diff
@@ -1713,9 +1713,12 @@ template IotaSlice(size_t N)
 
 // undocumented
 // zero cost variant of `std.range.iota`
+///
 struct IotaMap()
```
wilzbach, line 1718: I don't remember exactly, but you wanted to document it instead of setting it to `private` or `protected`?

9il, line 1718: just undocumented

wilzbach, line 1718: Dscanner rightfully complains about that as you shouldn't have anything that is public and undocumented...

9il, line 1718: Turn it off)

wilzbach:
> Turn it off)

Imho it is one of the more useful switches as I found > 10 occurrences in the codebase of undocumented public methods. So I would like to keep that flag.
Can't we set it to `protected` or `private` as long as you don't want to expose `IotaMap` to the user?

9il:
> Can't we set it to protected or private as long as you don't want to expose IotaMap to the user?

OK, document it)

wilzbach:
> OK, document it)

like this or with more text?

9il:
> like this or with more text?

few words

source/mir/ndslice/selection.d
```diff
@@ -1711,11 +1711,15 @@ template IotaSlice(size_t N)
     alias IotaSlice = Slice!(N, IotaMap!());
 }
 
-// undocumented
-// zero cost variant of `std.range.iota`
+/++
+Zero cost variant of `std.range.iota`
```
9il: Zero cost variant of `std.range.iota` - this should be with `//` . Just add that this range is used for iotaSlice.

9il: Oh, left it, but with small change: 

```
Zero cost variant of `std.range.iota` for iotaSlice.
```

And make `std.range.iota` a link to iota

wilzbach:
updated - i used the new `REF_ALTTEXT` macro , which doesn't work yet with `bootDoc`, but works for Phobos.

9il:
> REF_ALTTEXT

What is this and would it work for Phobos?

wilzbach:
> What is this and would it work for Phobos?

Yep I recently found it https://github.com/dlang/phobos/pull/4079 - i think it was part of a move to deprecate a lot of old macros within Phobos, but this PR stalled somehow.
It allows to have an alternative text for references - do you know a better macro?

9il:
> It allows to have an alternative text for references - do you know a better macro?
>  Hide all checks

no

PR #156, opened 2016-04-18
create executable examples

wilzbach:
I just saw that the `median-filter` example that was included in `package.d` doesn't work as `imageformats` was removed.

I think it is a lot nicer to have them in a separate folder `examples` that we also can easily execute. I searched a bit and it turns out that `dub´ supports dependencies depending on the configuration :)

Btw code coverage also complained about the unittest as it wasn't executed.

You can build & run it with `dub -c median-filter` (for it's own CLI arguments you need to append the `--` to it) - maybe that's worth mentioning in the README?

Update: [Issue at Dub](https://github.com/dlang/dub/issues/819) to simply the setup of creating multiple executables.

### DIFF
--- a/dub.json
+++ b/dub.json
@@ -1,13 +1,23 @@
 {
-	"name": "mir",
-	"description": "Numeric library for Dlang",
-	"copyright": "2015-2016, Ilya Yaroshenko, DlangScience",
-	"authors": ["Ilya Yaroshenko"],
-	"license": "BSL-1.0",
-	"configurations": [
-		{
-			"name": "d_source_library",
-			"targetType": "sourceLibrary",
-		},
-	]
+    "name": "mir",
+    "description": "Numeric library for Dlang",
+    "copyright": "2015-2016, Ilya Yaroshenko, DlangScience",
+    "authors": ["Ilya Yaroshenko"],
+    "license": "BSL-1.0",
+    "configurations": [
+        {
+            "name": "d_source_library",
+            "targetType": "sourceLibrary",
+        },
+        {
+            "name": "median-filter",
+            "targetName": "median-filter",
+            "targetType": "executable",
+            "targetPath": "build",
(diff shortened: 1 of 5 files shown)


9il:
> doesn't work as imageformats was removed

this is dub package

wilzbach:
> this is dub package

Yep but you removed those essential lines - leaving a unittest that basically doesn't get executed.
Thus if you would like to keep that in the `package.d` - maybe putting it in the ddoc header?
But I think being able to directly execute it is nicer?

9il:
should be included into build scripts

wilzbach:
> should be included into build scripts

Have a look at the changes and the `dub.json` - as said `dub -c median-filter` builds & executes the script.

dub.json
```diff
@@ -9,5 +9,15 @@
 			"name": "d_source_library",
 			"targetType": "sourceLibrary",
 		},
-	]
+        {
```
9il, line 12: small intendation

9il:
> Have a look at the changes and the dub.json - as said dub -c median-filter builds & executes the script.

We should build it at least in Circle CI

dub.json
wilzbach, line 12: hmm previously tabs were used and I used spaces - can I retab the entire file?

wilzbach:
> We should build it at least in Circle CI

added - should we also provide sample input and output?

dub.json
```diff
@@ -1,13 +1,23 @@
...
-	"copyright": "2015-2016, Ilya Yaroshenko, DlangScience",
-	"authors": ["Ilya Yaroshenko"],
-	"license": "BSL-1.0",
-	"configurations": [
-		{
-			"name": "d_source_library",
-			"targetType": "sourceLibrary",
-		},
-	]
+    "name": "mir",
```
wilzbach, line 2: replaced tabs with spaces

PR #161, opened 2016-04-18
improve coverage for mir.ndslice.slice

wilzbach:
Let's increase the coverage of ndslice to 100% :)

### DIFF
--- a/source/mir/ndslice/slice.d
+++ b/source/mir/ndslice/slice.d
@@ -1720,6 +1720,24 @@ struct Slice(size_t _N, _Range)
         assert(slice(indexP) == 4);  // Math & Fortran order
     }
 
+    static if (doUnittest)
+    pure nothrow unittest
+    {
+        // check with different PureN
+        import mir.ndslice.selection: pack, iotaSlice;
+        auto pElements = iotaSlice(2, 3, 4, 5).pack!2;
+        import std.range: iota;
+        import std.algorithm.comparison: equal;
+
+        // D & C order
+        assert(pElements[$-1, $-1][$-1].equal([5].iotaSlice(115)));
+        assert(pElements[[1, 2]][$-1].equal([5].iotaSlice(115)));
+
+        // Math & Fortran
+        assert(pElements(2, 1)[$-1].equal([5].iotaSlice(115)));
+        assert(pElements([2, 1])[$-1].equal([5].iotaSlice(115)));
+    }
+
     /++
     $(BLUE Partially or fully defined slice).
     +/
@@ -2755,6 +2773,12 @@ static if(__VERSION__ >= 2071)
         assert(ptr[0] == save0 + 11);
         (ptr + 5)[2] = 333;
         assert(range[7] == 333);
+
+        auto ptrCopy = ptr.save;
+        ptrCopy._range.popFront;
+        ptr[1] = 2;
+        assert(ptr[0] == save0 + 11);
+        assert(ptrCopy[0] == 2);
     }
 }
 
@@ -2776,6 +2800,17 @@ pure nothrow unittest
         slice[5 .. $][2] = 333;
         assert(range[7] == 333);
     }
+}
+
+unittest
+{
+    int[] arr = [1, 2, 3];
+    auto ptr = arr.ptrShell;
+    assert(ptr[0] == 1);
+    auto ptrCopy = ptr.save;
+    ptrCopy._range.popFront;
+    assert(ptr[0] == 1);
+    assert(ptrCopy[0] == 2);
 }
 
 private enum isSlicePointer(T) = isPointer!T || is(T : PtrShell!R, R);
@@ -2855,7 +2890,6 @@ private template PtrTuple(Names...)
 pure nothrow unittest
 {
     auto a = new int[20], b = new int[20];
-    import std.stdio;
     alias T = PtrTuple!("a", "b");
     alias S = T!(int*, int*);
     auto t = S(a.ptr, b.ptr);


source/mir/ndslice/slice.d
```diff
@@ -1144,17 +1144,15 @@ struct Slice(size_t _N, _Range)
         mixin(mathIndexStrideCode);
     }
 
+    static if (hasPtrBehavior!PureRange)
```
wilzbach: afaict this constructor is only called with pointers?

source/mir/ndslice/slice.d
```diff
@@ -2778,6 +2794,17 @@ pure nothrow unittest
...
 
+unittest
+{
+    int[] arr = [3, 4, 5, 6, 7, 8];
+    auto ptr = arr.ptrShell;
+    auto ptrCopy = ptr.save;
+    ptrCopy[2] = 1;
+    ptrCopy._range.popFront;
+    assert(ptr[0] == 3);
+    assert(ptrCopy[0] == 4);
```
wilzbach: is this a correct test for the `save` behavior?

source/mir/ndslice/slice.d
9il: no

source/mir/ndslice/slice.d
```diff
@@ -1144,17 +1144,15 @@ struct Slice(size_t _N, _Range)
...
         foreach (i; Iota!(0, PureN))
             _lengths[i] = lengths[i];
         foreach (i; Iota!(0, PureN))
             _strides[i] = strides[i];
-        static if (hasPtrBehavior!PureRange)
-            _ptr = range;
-        else
-            _ptr._range = range;
 
+        _ptr = range;
```
9il: but this is ok

source/mir/ndslice/slice.d
```diff
@@ -1720,6 +1718,24 @@ struct Slice(size_t _N, _Range)
         assert(slice(indexP) == 4);  // Math & Fortran order
     }
 
+
```
9il: remove line please

source/mir/ndslice/slice.d
```diff
@@ -1720,6 +1718,24 @@ struct Slice(size_t _N, _Range)
...
+    static if (doUnittest) unittest
```
9il: please use different lines

source/mir/ndslice/slice.d
9il: no, see dummyranges module in phobos sources

9il: You should define a range based on class representation

source/mir/ndslice/slice.d
```diff
@@ -2869,6 +2893,24 @@ pure nothrow unittest
     assert(b[0] == 2);
 }
 
+unittest
+{
+    // test save behavior
```
9il: the same, this test does not actually test save.

source/mir/ndslice/slice.d
wilzbach: doesn't work without the restriction

```
 cannot implicitly convert expression (range) of type Result to PtrShell!(Result)
```

9il: oh, understand now. Please don' fix this constructor. This is for internal use only

wilzbach: coverage says it is not used at all :/

9il: This is just not covered, but this is used by package

source/mir/ndslice/slice.d
```diff
@@ -2867,6 +2886,12 @@ pure nothrow unittest
     r = t.front;
     r.b = r.a * 2;
     assert(b[0] == 2);
+
+    // test save
+    auto c = new int[20];
+    auto tCopy = t.save;
+    tCopy.ptrs[0] = c.ptr;
+    assert(t.front.a == 1);
```
wilzbach: using dummy ranges didn't work - see #163

source/mir/ndslice/slice.d
wilzbach: > This is just not covered, but this is used by package

where?

source/mir/ndslice/slice.d
9il: this is wrong test

source/mir/ndslice/slice.d
9il: > where?

The line with `_ptr` is used, so the constructor is used.

wilzbach: maybe we are talking about different things, this constructor is always called with ranges that have `PtrBehavior`, so this line:

```
_ptr._range = range;
```

is never called.

9il: This line is never called for current test, but may be called for user code

9il: Anyway, this is correct line and i don't what this be removed.

source/mir/ndslice/slice.d
```diff
@@ -1144,27 +1144,15 @@ struct Slice(size_t _N, _Range)
         mixin(mathIndexStrideCode);
     }
 
-    this(ref in size_t[PureN] lengths, ref in sizediff_t[PureN] strides, PureRange range)
+    this(S)(ref in size_t[PureN] lengths, ref in sizediff_t[PureN] strides, S range)
+    if(hasPtrBehavior!PureRange || is(PtrShell!PureRange == S))
```
wilzbach: turns out that `_ptr` can only be either of them - see above:

```
    static if (hasPtrBehavior!PureRange)
        PureRange _ptr;
    else
        PtrShell!PureRange _ptr;
```

source/mir/ndslice/slice.d
```diff
@@ -2778,6 +2790,18 @@ pure nothrow unittest
     }
 }
 
+unittest
+{
+    // test save with dynamic arrays
```
wilzbach: dynamic arrays are handled differently in contrast to ranges, thus the extra test.

source/mir/ndslice/slice.d
9il: Please stop breaking logic of this constructor

9il: I need this logic in this module

9il: I am already spent a lot of time because you want to cover 1 line

source/mir/ndslice/slice.d
```diff
@@ -2810,8 +2834,6 @@ private template PtrTuple(Names...)
             foreach (i, ref ptr; ptrs)
                 static if (isPointer!(Ptrs[i]))
                     p.ptrs[i] = ptr;
-                else
-                    p.ptrs[i] = ptr.save;
```
9il, line 2814: Why????

9il:
If you can not cover something, do not remove it. Let it be uncovered please

wilzbach:
> I am already spent a lot of time because you want to cover 1 line

will rebase, s.t. that only the math test is in there and throw away the other stuff.

wilzbach:
> If you can not cover something, do not remove it. Let it be uncovered please

rebased.

PR #64, opened 2016-04-11
las: improve coverage

wilzbach:
Should increase the coverage by quite a bit.

PR #162, opened 2016-04-18
improve coverage for mir.las.sum

wilzbach:
Just adds 

@9il - the remaining [uncovered bits](https://codecov.io/github/DlangScience/mir/source/mir/las/sum.d) here beyond my knowledge - do you mind to have a look?
- L76
- L757-759
- L763-765
- L1744, L1750, L1756 (I already tried to tackle this bit in #64 - but you reverted due to failure on Windows?)

### DIFF
--- a/source/mir/las/sum.d
+++ b/source/mir/las/sum.d
@@ -1714,6 +1714,24 @@ public:
         }
     }
 
+    @nogc nothrow unittest
+    {
+        import std.typetuple;
+        import std.math: approxEqual;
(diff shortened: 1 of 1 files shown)


9il:
> do you mind to have a look?

Yes, but i don't want to spend time to cover second level of KB2 summation

9il:
> but you reverted due to failure on Windows

they was not correct

wilzbach:
> they was not correct

I know :/

PR #165, opened 2016-04-19
build examples on all platforms

wilzbach:
As we use this third-party library, I don't know whether it builds on Windows, but there is only one way to find out ;-)


wilzbach:
Wohoo it does - @9il feel free to merge ;-)

9il:
Great! We also need to add this to travis.

wilzbach:
> Great! We also need to add this to travis.

Already did - see
https://github.com/DlangScience/mir/pull/165/files

9il:
Cool!

PR #186, opened 2016-04-21
how to run guide for dmd and ldc

wilzbach:
Hey I added instruction how one can use mir with different compilers:

1) directly (source inclusion)
2) static lib
3) dynamic lib

I hope this is useful! I put the items in order of usefulness - that's why `rdmd` is on top, followed by `ldc`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,7 +8,8 @@ Mir
 ======
 Generic Numerical Library for Science and Machine Learning.
 
-## Contents
+Contents
+--------
 
  - `mir.ndslice` [Multidimensional Random Access Ranges and Arrays](http://dlang.org/phobos-prerelease/std_experimental_ndslice.html)
  - `mir.las.sum` Functions and Output Ranges for Summation Algorithms. Works with user-defined types.
@@ -22,11 +23,13 @@ Generic Numerical Library for Science and Machine Learning.
  - `mir.sparse` Sparse Tensors (see sparse branch and `v0.15.1-beta2`+)
   -  `mir.sparse.blas` Spars BLAS
 
-## Documentation
+Documentation
+-------------
 
 Alpha version of API is available [here](http://docs.mir.dlang.io/latest/index.html).
 
-## Compatibility
+Compatibility
+-------------
 
 |           | Linux | Mac OS X | Windows |
 |-----------|-------|----------|---------|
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -38,7 +41,88 @@ Alpha version of API is available [here](http://docs.mir.dlang.io/latest/index.h
...
+`rdmd` is a companion to the `dmd` compiler that simplifies the typical edit-compile-link-run or edit-make-run cycle to a rapid edit-run cycle. Like make and other tools, `rdmd` uses the relative dates of the files involved to minimize the amount of work necessary. Unlike make, `rdmd` tracks dependencies and freshness without requiring additional information from the user.
+You can find more information [here](https://dlang.org/rdmd.html).
+
+Compile with ldc
+----------------
+
+The easiest way is to specify mir's sources during compilation.
+
+```
+ldc -Isource examples/means_of_columns.d
```
wilzbach, line 97: In case you ask, the following doesn't work with `dmd`

```
dmd -Isource examples/means_of_columns.d
```

README.md
```diff
@@ -38,7 +41,88 @@ Alpha version of API is available [here](http://docs.mir.dlang.io/latest/index.h
 - DMD (reference D compiler) >= `2.068`
 - LDC (LLVM D Compiler) `0.17.0`+.
 
-## Packages
+Rapid edit-run cycle
```
9il: Please rename it to `Rapid edit-run cycle without dub`.

wilzbach: done.

README.md
```diff
@@ -38,7 +41,88 @@ Alpha version of API is available [here](http://docs.mir.dlang.io/latest/index.h
...
+
+#### 2.b.2: Link dynamically
+
+```
+dmd means_of_columns.o -Llibmir.so -L-rpath=.
+```
+
+You can inspect the dynamic linking with `ldd means_of_columns`.
 
 ### On any platform with dub
```
9il: This should be above on all notes about compiling and linking. Please make it clear for user, that if he has `dub` he does not need to read the notes about compilation. Scientific and math folk are very afraid any about problems with compilers

README.md
```diff
@@ -69,27 +154,32 @@ Now you can create a main file in the `source` and run your code with:
...
 | Stable | `libmir`     | [![libmir](https://img.shields.io/aur/version/libmir.svg)](https://aur.archlinux.org/packages/libmir/) [![libmir](https://img.shields.io/aur/votes/libmir.svg)](https://aur.archlinux.org/packages/libmir/) |
 | Latest | `libmir-git` | [![libmir-git](https://img.shields.io/aur/version/libmir-git.svg)](https://aur.archlinux.org/packages/libmir-git/) |
 
-### On any platform with import paths
-
-Mir is a pure source code library, that means it can be easily distributed to
-any system. So you can just copy Mir's source to your system's dlang import path.
-For example on  Linux this is `/usr/include/dlang/dmd/mir`
-(or `/usr/include/dlang/dmd/mir` for ldc).
+Other tips
```
9il: We do not need this section because mir has very tiny name space for linking. Furthermore, please never refer from Mir related pages to articles about vibe.d (it is very buggy library and user may get bad experience )

README.md
wilzbach: I could put it between `ldc` and `dmd`?
How about "On any play platform using the dub build tool"?

9il:
Also we need dub configurations to build static and dynamic libraries. source library is by default anyway

README.md
9il: > I could put it between ldc and dmd?

Why between `ldc` and `dmd` ? Dub should be on top .

> How about "On any play platform using the dub build tool"?

`Fast setup with DUB package manager` or something like that

wilzbach: > Why between ldc and dmd ? Dub should be on top .

because unfortunately it's not that fast, so I definitely prefer `rdmd` over `dub`.

9il:
Please do not use H1 sections for each compiler. All installation section section should be subsections of some kind of `Installation`

README.md
wilzbach: Understood & removed ;-)

wilzbach:
> Also we need dub configurations to build static and dynamic libraries. source library is by default anyway

Added.

```
dub build -c static-lib
dub build -c dynamic-lib
```

> Please do not use H1 sections for each compiler. All installation section section should be subsections of some kind of Installation

`----` is `h2`, but understood your point ;-)

README.md
```diff
@@ -38,9 +41,21 @@ Alpha version of API is available [here](http://docs.mir.dlang.io/latest/index.h
...
+The easiest way to execute your code is with `rdmd`.
+
+```
+rdmd -Isource examples/means_of_columns.d
+```
+
+`rdmd` is a companion to the `dmd` compiler that simplifies the typical edit-compile-link-run or edit-make-run cycle to a rapid edit-run cycle. Like make and other tools, `rdmd` uses the relative dates of the files involved to minimize the amount of work necessary. Unlike make, `rdmd` tracks dependencies and freshness without requiring additional information from the user.
+You can find more information [here](https://dlang.org/rdmd.html).
+
+### Fast setup with the package manager dub
```
9il: Please move to the first position =)

9il: words order `the dub package manager`

9il:
Thanks!

wilzbach:
> Thanks!

As always - you are very welcome ;-)

PR #190, opened 2016-04-24
add math doc module

9il:


wilzbach:
Btw I really liked the alias to llvm intrinsics - is there any chance we can get that into phobos? :)

9il:
> Btw I really liked the alias to llvm intrinsics - is there any chance we can get that into phobos? :)

They are, partially. But first we need `log`/`exp` for double and float numbers.

wilzbach:
Go for it :)

PR #192, opened 2016-04-24
remove tabs and insert check

wilzbach:
@9il I replaced all tabs with four spaces and removed the trailing whitespace. Moreover I inserted a check in our circle CI setup that checks for trailing whitespace and tabs for future additions ;-)


9il:
thanks!

PR #196, opened 2016-04-24
remove commented code

wilzbach:
we probably shouldn't have this commented code in the master branch?

### DIFF
--- a/source/mir/internal/utility.d
+++ b/source/mir/internal/utility.d
@@ -39,19 +39,3 @@ alias ConstData(S : T[], T) = const(T)[];
 
 enum isVector(V) = is(V : T1[], T1) || is(V : Slice!(1, T2*), T2);
 enum isMatrix(M) = is(M : Slice!(2, T*), T);
-
-//enum isConstVector(T) ==
-//    is(Slice!(1,           T [])) ||
-//    is(Slice!(1, const    (T)[])) ||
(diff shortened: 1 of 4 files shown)


source/mir/sparse/blas/gemm.d
```diff
@@ -197,11 +197,6 @@ unittest
          [-1.0, 9, 4, 8],
          [9.0, 8, 3, -2]];
 
-    // a * b ==
```
9il, line 200: This is comment for user

source/mir/internal/utility.d
```diff
@@ -39,19 +39,3 @@ alias ConstData(S : T[], T) = const(T)[];
 
 enum isVector(V) = is(V : T1[], T1) || is(V : Slice!(1, T2*), T2);
 enum isMatrix(M) = is(M : Slice!(2, T*), T);
-
-//enum isConstVector(T) ==
```
9il, line 43: Yes, this can be removed

source/mir/sparse/blas/dot.d
```diff
@@ -83,8 +83,6 @@ D dot(
         }
         continue;
         L:
-        //import mir.internal.math: fmuladd;
```
9il, line 86: yes, can be removed

wilzbach:
> This is comment for user

Oh I was too late - sorry. Rebased!

PR #197, opened 2016-04-25
Add exception msg for sum

9il:


source/mir/sum.d
```diff
@@ -1838,22 +1838,31 @@ nothrow unittest
...
-        else
+        catch(AssertError e)
         {
-            assert(summator.isNaN() == r.isNaN());
-            assert(summator.isFinite() == r.isFinite());
-            assert(summator.isInfinity() == r.isInfinity());
-            assert(s == r || s.isNaN && r.isNaN);
+            import std.stdio;
+            stderr.writefln("i = %s, result = %s (%X), target = %s (%X)", i, s, s, r, r);
         }
```
wilzbach, line 1866: Isn't stdio in a unittest bad or do you want to debug something specific?

9il, line 1866: > debug something specific

Yes

wilzbach, line 1866: Then go for it, but don't forget to remove it afterwards cause currently the test would pass with this error 
;-)

9il, line 1866: This is very buggy part for compilers

wilzbach, line 1866: > This is very buggy part for compilers

Yep but that's exactly why we should test for it

PR #198, opened 2016-04-25
test mir with ldc-1.0.0-beta1

wilzbach:
let's test whether ldc-1.0.0-beta1 works for us ;-)
(-> please wait until all CIs pass)

It's based on 2.070.2, so maybe we soon don't need all the `version` code for the allocator anymore ;-)

http://forum.dlang.org/thread/vnuknfqnlzzlxhzyfakc@forum.dlang.org


9il:
revert this?

wilzbach:
yep :/

PR #199, opened 2016-04-25
Update .travis.yml

9il:


wilzbach:
how about making it optional?
Then it's only downloaded for ldc.
You need to remove `ldc` 1.0.0 from the matrix then & I forgot to update the x86 target.

```
    - {os: linux, d: ldc-1.0.0-beta1, addons: {apt: {packages: [[libconfig++8]]}}}
```

9il:
This PR does not help

PR #203, opened 2016-05-02
use dlang.org's ddoc + theme

wilzbach:
Should finally solve our documentation problems.
In custom.ddoc we can overwrite dlang.org's macros - we might want to do that for links going to Phobos.
Apart from that everything else is copied from dlang.org, so we should be able to test better our macros ;-)

Also it looks pretty nice: http://docs.mir.dlang.io/ddoc/ 

1) I set circle, s.t. it will build the new html once you merge this
2) You can preview locally with `make -f doc/Makefile html`
3) I tried to clean the Makefile as far as I could - I hope it's understandable
4) I used the modlist.d menu generator from dlang.org
5) I put the documentation stuff into a new `doc` folder, s.t. we aren't that messy on the root folder

Also: 
I) the overview page needs some work, but I guess we can do this once this is merged.
II)  I quickly changed the "branding", s.t. no one will complain. We probably want to select better colors in the future?


9il:
Awesome!!!

> We probably want to select better colors in the future?

Yes, we can change logo colors too)

PR #210, opened 2016-05-10
add gemm, gemv, dot

9il:
Code example: https://gist.github.com/9il/58c1b80110de2db5f2eff6999346a928


source/mir/blas/kernel/gemm.d
```diff
@@ -0,0 +1,91 @@
...
+
+enum Conj
+{
+	none,
+	complexNone,
+	complexA,
+	complexB,
+}
+
+void gemmMicroKernel(
```
wilzbach: Public declaration 'gemmMicroKernel' is undocumented

source/mir/blas/kernel/gemm.d
```diff
@@ -0,0 +1,91 @@
...
+	scope V[P][N][M]* c,
+	)
+{
+	pragma(inline, false);
+	enum msg = "Wrong kernel compile time arguments.";
+	static assert(conj == Conj.none && P == 1 || conj != Conj.none && P == 2, msg);
+	do
+	{
+		V[P][N][M] reg = void;
+		foreach(m; Iota!(0, M))
```
wilzbach: whitespace ;-)

source/mir/blas/kernel/gemm.d
```diff
@@ -0,0 +1,91 @@
+module mir.blas.generic.gemm;
+
+import mir.internal.utility;
+
+import std.typecons: Flag;
+
+enum Conj
+{
+	none,
```
wilzbach: tabs ...

source/mir/blas/kernel/gemm.d
```diff
@@ -0,0 +1,91 @@
...
+	scope const(F[P][M])* b,
+	scope V[P][N][M]* c,
+	)
+{
+	pragma(inline, false);
+	enum msg = "Wrong kernel compile time arguments.";
+	static assert(conj == Conj.none && P == 1 || conj != Conj.none && P == 2, msg);
+	do
+	{
+		V[P][N][M] reg = void;
```
wilzbach: does this really work? why do you need the `void` assignment?

source/mir/blas/kernel/gemm.d
```diff
@@ -0,0 +1,91 @@
...
+				reg[m][n][p] = c[0][m][n][p];
+			else
+				reg[m][n][p] = 0;
+		size_t i = kc;
+		do
+		{
+			V[P][N] ai = void;
+			V[P][M] bi = void;
+			foreach(n; Iota!(0, N))
+			foreach(p; Iota!(0, P))
```
wilzbach: nitpick: indent after loop statements?

source/mir/blas/kernel/gemm.d
```diff
@@ -0,0 +1,91 @@
...
+				else static assert(conj == Conj.none, msg);
+			}
+		}
+		while(--i);
+		b -= kc;
+		foreach(m; Iota!(0, M))
+		foreach(n; Iota!(0, N))
+		foreach(p; Iota!(0, P))
+			c[0][m][n][p] = reg[m][n][p];
+		c++;
```
wilzbach: shouldn't that yield a compile error? afaik you can't increment arrays

source/mir/blas/kernel/gemm.d
```diff
@@ -0,0 +1,91 @@
...
+	size_t P,
+	size_t N,
+	size_t M,
+	V,
+	F
+)(
+	in size_t kc,
+	size_t blocks,
+	scope const(V[P][N])* a,
+	scope const(F[P][M])* b,
```
wilzbach: does `scope` and `const` make sense together?

> This means that the destructor for an object is automatically called when the reference to it goes out of scope. The destructor is called even if the scope is exited via a thrown exception, thus scope is used to guarantee cleanup.
> ([scope](https://dlang.org/spec/attribute.html#scope))

source/mir/blas/kernel/gemm.d
```diff
@@ -0,0 +1,91 @@
...
+				}
+				else static assert(conj == Conj.none, msg);
+			}
+		}
+		while(--i);
+		b -= kc;
+		foreach(m; Iota!(0, M))
+		foreach(n; Iota!(0, N))
+		foreach(p; Iota!(0, P))
+			c[0][m][n][p] = reg[m][n][p];
```
wilzbach: I don't know much about GPU programming, but the need to copy the entire array seems weird. Why don't you directly assign to `c`?
Btw `c` has only three dimensions: `V[P][N][M]` and in the other way around

wilzbach:
@9il 
1) we have now CircleCI which does all the nitpicking - have a look at
https://circleci.com/gh/libmir/mir/170
2) Is there any chance you could provide a unittest? My "virtual" compiler tells me that this shouldn't compile.

source/mir/blas/kernel/gemm.d
9il: c is a pointer

source/mir/blas/kernel/gemm.d
9il: Yes, const refers to the data, scope to referece

source/mir/blas/kernel/gemm.d
9il: `c` is array of 3D static arrays :-)

9il:
> 2) Is there any chance you could provide a unittest? My "virtual" compiler tells me that this shouldn't compile.

This compiles =)

source/mir/blas/kernel/gemm.d
9il: for the last commit

source/mir/blas/kernel/gemm.d
9il: Yes, void assignment disallow default initilization

source/mir/blas/kernel/gemm.d
9il: all foreach loops are compile time, and without addition indentation code looks clear

source/mir/blas/kernel/gemm.d
9il: > Why don't you directly assign to c?

Directly assign may allow compiler to call `memcpy`

source/mir/blas/kernel/gemm.d
```diff
@@ -0,0 +1,151 @@
...
+    foreach (add; AliasSeq!(No.add, Yes.add))
+    {
+        enum P = conj == none ? 1 : 2;
+        {alias temp = gemmMicroKernel!(conj, add, P, 2 / P, 4 / P, float, float);}
+        {alias temp = gemmMicroKernel!(conj, add, P, 2 / P, 4 / P, double, double);}
+        version(LDC)
+        {
+            {alias temp = gemmMicroKernel!(conj, add, P, 2 / P, 4 / P, __vector(float[8]), float);}
+            {alias temp = gemmMicroKernel!(conj, add, P, 2 / P, 4 / P, __vector(double[4]), double);}
+        }
```
wilzbach: Can we have an assert statement here to verify the kernel code is actually correct?

9il: this is just a draft. Will be tested for final code

PR #217, opened 2016-05-16
update ldc to 1.0.0-beta2

wilzbach:
Seems like the ldc devs fixed the static linking error :)
https://github.com/ldc-developers/ldc/releases

@9il A bit of patience for travis this time please ;-)


wilzbach:
All checks have passed :)

PR #222, opened 2016-06-02
[WIP] Transformed density rejection with inflection points

wilzbach:
This is my WIP Tinflex branch that tracks my progress.
Paper: http://epub.wu.ac.at/3158/1/techreport-110.pdf 

What is working:
- concave/convex type determination for any interval with at most one inflection point
- computing hat and squeeze function for any interval with at most one inflection point

What's missing:
- compute x0 (defined as the maximum within the interval - I still need to figure out how they compute this efficiently)
- compute areas (I need to find out how "c" is selected)
- split intervals (arc-mean formula)
- sample new random value and filter

Please don't make the effort to criticize the style yet - it's WIP.


source/mir/random/generic.d
```diff
@@ -0,0 +1,384 @@
...
+    alias ti = tinflexInternal!(f0, f1, f2);
+
+    with(T1Type)
+    {
+        // type 1a: concave
+        assert(ti.determineType(0, 2 * PI) == T1a);
+        assert(ti.determineType(2 * PI, 4 * PI) == T1a);
+        assert(ti.determineType(2., 4) == T1a);
+        assert(ti.determineType(0., 5) == T1a);
+        assert(ti.determineType(1., 5) == T1a);
```
joseph-wakeling-sociomantic: Stylistic suggestion: always use an explicit `n.0` for floating-point, rather than this notation.  The characters/keystrokes saved are not worth it for the cost in readability/clarity of meaning.

source/mir/random/generic.d
```diff
@@ -0,0 +1,384 @@
...
+
+Authors: Sebastian Wilzbach, Ilya Yaroshenko
+
+TODO:
+- automatic method to identify inflection points?
+
+*/
+module mir.random.generic;
+
+alias f_t = double;
```
joseph-wakeling-sociomantic: Is this used anywhere in the module?  Can you explain the need?

source/mir/random/generic.d
```diff
@@ -0,0 +1,384 @@
...
+
+import std.stdio;
```
joseph-wakeling-sociomantic: Would suggest moving this to the local place where you use it (because you'll probably want to delete it together with any `writeln` statements in the final version).

source/mir/random/generic.d
```diff
@@ -0,0 +1,384 @@
...
+auto tinflex(T...)(T bs)
+{
+    import std.functional: unaryFun;
+    auto g0 = unaryFun!f0;
+    auto g1 = unaryFun!f1;
+    auto g2 = unaryFun!f2;
+    return tinflexInternal!(g0, g1, g2)(bs);
+}
+
+enum T1Type {T1a, T1b, T2a, T2b, T3a, T3b, T4a, T4b}
```
joseph-wakeling-sociomantic: This is very, VERY opaque in its meaning and purpose.  Unless there's a very, very good reason (e.g. matching up with notation in the Tinflex paper) I would try to give both the type and the individual values more semantically meaningful names.

If there is a logic to the naming choices, then it should be documented, because at first glance, it looks arbitrary.

source/mir/random/generic.d
```diff
@@ -0,0 +1,384 @@
...
+    $(LI $(F_TILDE) is entirely concave (`f'' is entirely negative))
+    $(LI $(F_TILDE) is entirely convex (`f'' is entirely positive))
+    $(LI $(F_TILDE) contains one inflection point (`f'' intersects the x-axis once)
+)
+
+Macros:
+    F_TILDE=g
+*/
+
+auto tinflex(T...)(T bs)
```
joseph-wakeling-sociomantic: I would recommend that you try to pick semantically meaningful (i.e. self-documenting) names for your parameters as a matter of habit.

In cases where you're trying to match notation in the Tinflex paper, give consideration to when that's really important; if you feel the match is important, document the meaning clearly.

source/mir/random/generic.d
```diff
@@ -0,0 +1,384 @@
...
+*/
+bool isConvex(T1Type t)
+{
+    return !isConcave(t);
+}
+
+/**
+
+*/
+template tinflexInternal(alias f0, alias f1, alias f2)
```
joseph-wakeling-sociomantic: Assuming `f0`, `f1` and `f2` are meant to be callable functions, you could validate this using `std.traits.isCallable`.

source/mir/random/generic.d
```diff
@@ -0,0 +1,384 @@
...
+
+enum T1Type {T1a, T1b, T2a, T2b, T3a, T3b, T4a, T4b}
+
+/**
+Checks whether our function type is concave
+*/
+bool isConcave(T1Type t)
+{
+    with(T1Type)
+    return t == T1a || t == T2a || T3a || t == T4a;
```
joseph-wakeling-sociomantic: Note how difficult it is to understand the logic of these choices, and hence the behaviour of the function, without an obvious semantic meaning to the names of the enum values.

joseph-wakeling-sociomantic:
> Please don't make the effort to criticize the style yet - it's WIP.

It may seem fussy, but I think that getting style right from the start is an important consideration.  It makes things much easier to review for _technical_ content, if the style helps us to follow the code more easily.

A few pieces of advice in this respect.

First, I would suggest introducing basic documentation for _all_ your code features.  Every function should have a brief description, plus `Params:` and `Returns:` sections (and `Template_Params:` where appropriate), and similarly for data structures, templates, etc.

This will help us (as reviewers) to understand your intentions more easily, and will help you in terms of helping to clarify your thought on how the pieces of functionality fit together.

Second, I would suggest trying to use semantically meaningful parameter, variable and type names to the greatest extent possible.  If you want to maintain naming in line with the Tinflex paper, document the meaning of those names.

Any time you put into this will repay itself many times over both in speeding up effective technical review and in terms of the long-term maintainability of your code.

Making this a second-nature habit will also stand you in good stead if you ever find yourself wanting to apply for a software developer job anywhere ;-)

9il:
> It may seem fussy, but I think that getting style right from the start is an important consideration. It makes things much easier to review for technical content, if the style helps us to follow the code more easily.

This is not ready to review module. This is work in progress PR. @wilzbach should not fix this code now. First the module should be ready for review and then we can start comment about style, names and ask questions. Otherwise progress will be very slow.

9il:
> First, I would suggest introducing basic documentation for all your code features. Every function should have a brief description, plus Params: and Returns: sections (and Template_Params: where appropriate), and similarly for data structures, templates, etc.

Detailed documentation is the last part of review process. Please do not force @wilzbach to write it now

9il:
> Second, I would suggest trying to use semantically meaningful parameter, variable and type names to the greatest extent possible. If you want to maintain naming in line with the Tinflex paper, document the meaning of those names.

Encoded names is common practice for math code. Again, the docs is the last iteration

joseph-wakeling-sociomantic:
> This is not ready to review module. This is work in progress PR.

@9il obviously, your project, your review conventions.  I would simply ask you to consider that I don't make these suggestions arbitrarily.

My experience, based on working in a production environment, is that even WiP code should try to respect certain stylistic bottom lines.  Making this a habit saves time for everyone by minimizing the opportunities for misunderstanding or confusion.

> Encoded names is common practice for math code.

I'm familiar with typical habits of maths code.  That it's common practice, doesn't necessarily make it _good_ practice -- a lot of maths code is unnecessarily hard to read and refactor.

Ensuring that every parameter has an unambiguous meaning, either by its name or via documentation, is _very_ cheap to do, and makes it much easier to offer useful _technical_ feedback from the earliest possible stage.

joseph-wakeling-sociomantic:
Oh, and note, I can't force @wilzbach or anyone else to do anything.  I'm offering a point of view, which may be taken into account or discarded. ;-)

source/mir/random/generic.d
wilzbach: it was a hack to avoid the need to use which floating point - luckily I didn't need to choice it.

source/mir/random/generic.d
wilzbach: it's very convenient to have this in there for debugging - I can just insert writeln calls anymore. If you don't mind I will leave this in for a bit?

source/mir/random/generic.d
wilzbach: a: concave
b: convex

> Unless there's a very, very good reason (e.g. matching up with notation in the Tinflex paper

Yes :/

source/mir/random/generic.d
wilzbach: thanks!

wilzbach:
Travis CI fails because such nice aliasing for functions is only possible with D >= 2.070.

```
alias f0 = (x) => x ^^ 3;
alias f1 = (x) => 3 * x ^^ 2;
alias f2 = (x) => 6 * x;
```

@9il as this is not planned to be merged soon, at what point can we drop support for the older interfaces?
I heard that a new release for both dmd and ldc is coming up soon.

joseph-wakeling-sociomantic:
> Travis CI fails because such nice aliasing for functions is only possible with D >= 2.070.

Personally, I don't think the niceness of the notation justifies lack of backwards compatibility.  Breaking change is best reserved for features that really make a difference ;-)

source/mir/random/generic.d
wilzbach: btw this is not supposed to be exposed to any user

9il:
> @9il as this is not planned to be merged soon, at what point can we drop support for the older interfaces?

Yes

9il:
> Personally, I don't think the niceness of the notation justifies lack of backwards compatibility. Breaking change is best reserved for features that really make a difference ;-)

BLAS would break backward compatibility too because major reasons

joseph-wakeling-sociomantic:
> BLAS would break backward compatibility too because major reasons

Fair enough :-)

source/mir/random/generic.d
wilzbach: > Assuming f0, f1 and f2 are meant to be callable functions

turns out std.traits.isCallable doesn't work with alias function, putting this on hold for now.

source/mir/random/generic.d
wilzbach: I tried to explain it a bit more above,

Btw sorry for the confusion - this is an internal function.

wilzbach:
> Oh, and note, I can't force @wilzbach or anyone else to do anything. I'm offering a point of view, which may be taken into account or discarded. ;-)

Point taken. Having bad naming also makes it hard for myself to work on the code - that being said some naming comes from Tinflex, but I will put some explanations at the coments header.

source/mir/random/generic.d
joseph-wakeling-sociomantic: Yea, I phrased that badly.  I meant to suggest really that you try to make all your imports (including the ones needed for debugging) as local as possible, but your WiP use-case is fine, so carry on.

source/mir/random/generic.d
wilzbach: Alrighty - I will remove the Python/NumPy style.

source/mir/random/generic.d
dnadlinger: There seems to be an == missing from this, making it always true.

wilzbach: Thanks I already fixed that one, but didn't push it yet ;-)

wilzbach:
@9il @joseph-wakeling-sociomantic 

Just a quick update from my side:
- I have got the "initial" version of Tinflex working :tada: 
- The first parts (type determination, hat and squeeze function calculation, area between calculation) are rather "tested"
- The last part (random sampling based on the the produced areas) isn't "tested" yet (that's what I will do over the next two days)

-> now it's plotting time!

image

(more updates will follow in the Sunday update)

joseph-wakeling-sociomantic:
> I have got the "initial" version of Tinflex working

Congratulations!! :-D  I'll try to get you some feedback before the weekend.

joseph-wakeling-sociomantic:
> Travis CI fails because such nice aliasing for functions is only possible with D >= 2.070.
> 
> ``` D
> alias f0 = (x) => x ^^ 3;
> alias f1 = (x) => 3 * x ^^ 2;
> alias f2 = (x) => 6 * x;
> ```

Further remark on this, just for consideration.

While mir as a whole may require a more up to date compiler (e.g. because of BLAS), there's no reason I can see for making any one module require a more up to date compiler than is strictly needed.

Your downstream user may want to use only this one module or package, or may even want to copy-paste into another project.  It's a friendly and responsible thing to place minimal barriers in front of them doing this.  So, I would ask you to consider whether a minor convenience of notation should be put ahead of backwards compatibility.

wilzbach:
> Congratulations!! :-D I'll try to get you some feedback before the weekend.

Thanks! Please don't spend that much time yet - it's still pretty ugly!

joseph-wakeling-sociomantic:
> Please don't spend that much time yet - it's still pretty ugly!

Is there anything in particular you'd like me to give my attention to?

9il:
Please `FP ^^ FP`, where FP is Floating Point Type, with `pow` function from `mir.internal.math`. And `x ^^ 2` should be replaced with `x * x` (or `y = x`, `y * y` if x is expression) unless it is optimised to `x * x` even without optimisation flags.

wilzbach:
> Is there anything in particular you'd like me to give my attention to?

(I just wrote this because it's not ready yet for line-by-line nitpicking)

I will try to summarize the current concept & structure, so that it's hopefully easier to see global flaws and due to the sometimes missing comments probably desperately needed.

## Tinflex

Tinflex needs the following input:
- three functions (pdf and it's first two derivates)
- start intervals which can contain at most one inflection point
- target ratio rho (ratio between below hat and below squeeze area) = efficiency of the sampling
- parameter c (selection of the transformation function)

One can group the Tinflex algorithm in three parts
1) Setup of the initial intervals 
2) Splitting of the initial intervals until target efficiency is reached (generation loop)
3) Sampling based on the generated areas (acceptance-rejection loop)

I grouped the logic into the following modules

### Internal.d:  `LinearFun`

My simple abstraction of a linear function and to helper methods to create such for a secant or tangent.
The only tricky part here is that in the paper they define their functions with `f(x)=a + b*(x-y)`, whereas I thought just storing the slope and intercept would be enough `f(x) = bx + k`
(It turned out that some parts actually need the x and y of the point on which the tangent or secant was constructed - for secant higher one is picked.)

### Types.d
- `IntervalPoint` is the most important data construct. It's used to cache the values of the function (and it's derivatives), the area of hat and squeeze and stores a reference to it's the right neighbor. 
- `transform` and `transformtoInterval` were a bit difficult (I hit a compiler bug there) - the main problem was to return the three transformed functions 
- The other functions in `types.d` are more or less directly defined in the paper.

### Area.d
- hat and squeeze determination (as defined directly in the paper)
- `area` is the more interesting function that calculates the area below hat/squeeze

### Package.d
- splitting of intervals `calcPoints`
- acceptance-rejection sampling based on those intervals (as said, this part still has to be checked)

source/mir/random/generic/types.d
```diff
@@ -0,0 +1,365 @@
...
+*/
+public enum FunType {T1a, T1b, T2a, T2b, T3a, T3b, T4a, T4b}
+
+/**
+Checks whether our function type is concave
+*/
+protected bool isConcave(FunType t)
+{
+    with(FunType)
+    return t == T1a || t == T2a || t == T3a || t == T4a;
```
9il: `return t & 1 == 0`;

9il:
Q1: Does Tinflex algorithm require `FunType` declaration or the `switch` can be eliminated?
Q2: Do you want low level API be available for end user? reasons?

Important API issue: API should have minimalistic template bloat. Cost of calling `function` or `delegate` is small comparing with other required calculations. Please avoid any `alias` declarations for fX. Something like
`aDeclaration(F0, F1, F2)(scope F0 f0, scope F1 f1, scope F2 f2)` can be used.

9il:
And ping me please when I can start line-by-line review

wilzbach:
> Q1: Does Tinflex algorithm require FunType declaration or the switch can be eliminated?

It is internally required, but can be automatically detected.

> Q2: Do you want low level API be available for end user? reasons?

I don't see any. This was just not cleaned up ;-)

> Important API issue: API should have minimalistic template bloat. Cost of calling function or delegate is small comparing with other required calculations. Please avoid any alias declarations for fX. 

Wow that was a very helpful hint as I was always fighting against the compiler with the aliases.
The implementation is now `alias`-free (hopefully this removes our break in backwards compatibility) and I used `in` for e.g. the function, because we aren't going to modify them anyhow.

> And ping me please when I can start line-by-line review

Yes please.
I think there is still a minor issue with more complicated distributions, e.g. I am still trying to figure out why it's not sampling the normal distribution correctly :/
(more details via mail soon)

wilzbach:
I will start with the good news first. I added all [tests](https://github.com/cran/Tinflex/blob/master/tests/tests.R) from the Tinflex test suite and also similar plotting abilities, s.t. finding the last bugs gets easier.
For example
1) histogram and density function combined:

dist0_hist

2) The constructed interval functions
(blue is the upper hat function, whereas red is the lower squeeze function)

dist0_hs

(this can be executed with `dub -c generic-plot` and will create a folder `plots`).

Bad news:
- normal distribution still yields wrong results
- only 2 from six distributions of the Tinflex paper work (probably related to the above behavior)
- the testing with chi-squared tests is there, but the Tinflex paper depends on two other R packages (rvgtest and Runuran). From the latter an an implementation of  [pinv.new](http://www.inside-r.org/packages/cran/Runuran/docs/pinv.new) is missing (Polynomial interpolation of inverse CDF)

@9il Would you have time to start checking at least the proposed public interface in `package.d`? :)

9il:
@wilzbach Yes. Please share your normal distribution test and charts to see where it is wrong.

Also I am interesting how use should choose interval points. Do more point give better random numbers?

examples/plot_normal.d
```diff
@@ -0,0 +1,51 @@
+/**
+Simple plot
```
9il, line 2: OK, i see normal here)

examples/plot_normal.d
```diff
@@ -0,0 +1,51 @@
...
+    auto f1_below = sqrt(2 * π) * pow(σ, 3);
+    auto f1 = (real x) => -(x - μ) * exp(-xu2(x) / ss2) / f1_below;
+
+    // second derivative
+    auto f2_above = (real x) => exp(-xu2(x) / ss2);
+    auto f2_below_l = sqrt(2 * π) * ss2 * ss2 * σ;
+    auto f2_below_r = sqrt(2 * π) * ss2 * σ;
+    auto f2 = (real x) => (f2_above(x) * xu2(x)) / f2_below_l - (f2_above(x) / f2_below_r);
+
+    real[] points = [-1.0, 0.0, 1.0];
```
9il: are it correct points?

source/mir/analysis.d
```diff
@@ -0,0 +1,301 @@
+/**
+Statistical analysis functions
```
9il: Should be in some `inernal` package

source/mir/random/generic/package.d
```diff
@@ -0,0 +1,261 @@
...
+/**
+The Transformed Density Rejection with Inflection Points (Tinflex) algorithm
+can sample from arbitrary distributions given its density function f, its
+first two derivatives and a partitioning into intervals with at most one inflection
+point.
+
+Params:
+    f0 = probability density function of the distribution
+    f1 = first derivative of f0
+    f1 = second derivative of f0
```
9il: should be `f2`

source/mir/random/generic/package.d
```diff
@@ -0,0 +1,261 @@
...
+    c = T_c family
```
9il: link to the `T_c` description

source/mir/random/generic/package.d
```diff
@@ -0,0 +1,261 @@
...
+    points = non-overlapping partitioning with at most one inflection point per interval
+    rho = efficiency of the Tinflex algorithm
+
+Returns:
+    Tinflex Generator.
+*/
+Tinflex!(F0, S) tinflex(F0, F1, F2, S)
+               (in F0 f0, in F1 f1, in F2 f2,
```
9il: `scope` for callable objects and `in` for all others

source/mir/random/generic/package.d
```diff
@@ -0,0 +1,261 @@
...
+    f1 = first derivative of f0
+    f1 = second derivative of f0
+    c = T_c family
+    points = non-overlapping partitioning with at most one inflection point per interval
+    rho = efficiency of the Tinflex algorithm
+
+Returns:
+    Tinflex Generator.
+*/
+Tinflex!(F0, S) tinflex(F0, F1, F2, S)
```
9il: tinflex function should accept already calculated `IntervalPoint` too

source/mir/random/generic/package.d
```diff
@@ -0,0 +1,261 @@
...
+*/
+struct Tinflex(F0, S)
+{
+    // saved internal state of Tinflex:
+
+    // density function
+    private F0 _f0;
+
+    /// density function of the distribution
+    S f0(S x)
```
9il: should be `F0 pdf() const @property` instead.

source/mir/random/generic/package.d
```diff
@@ -0,0 +1,261 @@
...
+    {
+        return _f0(x);
+    }
+
+    // generated partition points
+    private IntervalPoint!S[] ips;
```
9il: move all declaration to the begining

9il: should be `const`

source/mir/random/generic/package.d
```diff
@@ -0,0 +1,261 @@
...
+
+    // global T_c family
+    private S c;
+
+    private this(F0 f0, IntervalPoint!S[] ips, S c)
```
9il: ~~`scope F0`~~ `const F0` and const `ips`

source/mir/random/generic/package.d
9il: hmm, ok, lets use `in` for callable function objects `opCall` methods are `const`

source/mir/random/generic/package.d
```diff
@@ -0,0 +1,261 @@
...
+/**
+Data body of the Tinflex algorithm.
+Can be used to sample from the distribution.
+*/
+struct Tinflex(F0, S)
+{
+    // saved internal state of Tinflex:
+
+    // density function
+    private F0 _f0;
```
9il: should be `const`

source/mir/random/generic/package.d
```diff
@@ -0,0 +1,261 @@
...
+
+    import std.random: dice, uniform;
+    import std.math: abs;
+    import mir.internal.math: exp;
+
+    import mir.random.generic.transformations : inverse, antiderivative, inverseAntiderivative;
+
+    double X;
+    // acceptance-rejection sampling
+    while (true)
```
9il: Phobos style is `for(;;)`

source/mir/random/generic/package.d
```diff
@@ -0,0 +1,261 @@
...
+
+        if (u * hatX <= squeezeX)
+        {
+            break;
+        }
+        else if (u * hatX <= exp(f0(X)))
+        {
+            break;
+        }
+
```
9il: empty line

source/mir/random/generic/package.d
```diff
@@ -0,0 +1,261 @@
...
+    import mir.random.generic.transformations : inverse, antiderivative, inverseAntiderivative;
+
+    double X;
+    // acceptance-rejection sampling
+    while (true)
+    {
+        auto j = dice(rng, areas);
+        auto u = uniform(0, 1, rng);
+
+        if (abs(ips[j].hat.slope) > 1e-10)
```
9il: Why this constant `1e-10`? What constraints for possible values of this constant?

source/mir/random/generic/package.d
```diff
@@ -0,0 +1,261 @@
...
+    import mir.internal.math: exp;
+
+    import mir.random.generic.transformations : inverse, antiderivative, inverseAntiderivative;
+
+    double X;
+    // acceptance-rejection sampling
+    while (true)
+    {
+        auto j = dice(rng, areas);
+        auto u = uniform(0, 1, rng);
```
9il: Why not `uniform01`?
Should be left bound included?
Should be right bound not included?

source/mir/random/generic/package.d
```diff
@@ -0,0 +1,261 @@
...
+            X = ips[j].hat._y + (inverseAntiderivative(antiderivative(ips[j].hat(ips[j].x), c)
+                          + ips[j].hat.slope * u, c) - ips[j].hat.a) / ips[j].hat.slope;
+        }
+        else
+        {
+            X = (1-u) * ips[j].x + u * ips[j + 1].x;
+        }
+
+        auto hatX = inverse(ips[j].hat(X), c);
+        double squeezeX;
```
9il: `double squeezeX ... ? ... : 0`

source/mir/random/generic/package.d
9il: Why `double`?

source/mir/random/generic/package.d
```diff
@@ -0,0 +1,261 @@
...
+/**
+Convenience method to sample Arrays with sample r
+This will be replaced with a more sophisticated version in later versions.
+
+Params:
+    r = random sampler
+    n = number of times to sample
+Returns: Randomly sampled Array of length n
+*/
+typeof(R.init())[] sample(R)(R r, int n)
```
9il: Should not be a part of this module

9il: Should be internal for now

source/mir/random/generic/package.d
```diff
@@ -0,0 +1,261 @@
+/**
```
9il: Should be `mir/random/tinflex` instead of `generic` because other generic approaches are possible

source/mir/random/generic/package.d
```diff
@@ -0,0 +1,261 @@
...
+            X = (1-u) * ips[j].x + u * ips[j + 1].x;
+        }
+
+        auto hatX = inverse(ips[j].hat(X), c);
+        double squeezeX;
+        if (ips[j].squeezeA > 0)
+            squeezeX = inverse(ips[j].squeeze(X), c);
+        else
+            squeezeX = 0;
+
```
9il: add `immutable t = u * hatX`;

source/mir/random/generic/package.d
```diff
@@ -0,0 +1,261 @@
...
+        if (u * hatX <= squeezeX)
+        {
+            break;
+        }
+        else if (u * hatX <= exp(f0(X)))
```
9il: does not need `else` here

source/mir/random/generic/package.d
```diff
@@ -0,0 +1,261 @@
...
+/**
+(For testing only - will be moved)
+Generates a series of y-values that can be used for plotting.
+
+Params:
+    t = Tinflex generator
+    xs = x points to be plotted
+    hat = whether hat (true) or squeeze (false) should be plotted
+*/
+auto plot(F0, S)(Tinflex!(F0, S) t, S[] xs, bool hat = true)
```
9il: should be internal for now

source/mir/random/generic/transformations.d
```diff
@@ -0,0 +1,147 @@
...
+Create a c-transformation, based on a function and it's first two derivatives
+*/
+auto transformToInterval(F0, F1, F2, S)(in F0 f0, in F1 f1, in F2 f2, in S c)
+    if (is(ReturnType!F0 == S) && is(ReturnType!F1 == S) && is(ReturnType!F2 == S))
+{
+    import std.math: sgn;
+    import mir.internal.math: pow, exp;
+    import mir.random.generic.types : intervalPoint, IntervalPoint;
+
+    // TODO: use caching
```
9il: Why caching?

source/mir/random/generic/transformations.d
```diff
@@ -0,0 +1,147 @@
...
+    import mir.internal.math : exp;
+    import std.math : sgn;
+    if (c == 0)
+        return exp(x);
+    else if (c == -0.5)
+        return 1 / x ^^ x;
+    else if (c == 1)
+        return x;
+    else
+        return (sgn(c) * x) ^^ (1 / c);
```
9il: can `copysing` be used here?

source/mir/random/generic/transformations.d
```diff
@@ -0,0 +1,147 @@
...
+    import mir.internal.math : exp, log;
+    import std.math : sgn;
+    if (c == 0)
+        return log(x);
+    else if (c == -0.5)
+        return -1 / x;
+    else if (c == -1)
+        return exp(x);
+    else
+        return sgn(c) * (sgn(c) * (c + 1) / c * x) ^^ (c / (c + 1));
```
9il: can `sgn(c) * (c + 1) / c * x)` be negative?
If yes, can be `(c / (c + 1))` not an integer?

source/mir/random/generic/types.d
```diff
@@ -0,0 +1,232 @@
...
+
+/**
+Major data unit of the Tinflex algorithm.
+It is used to store
+- (cached) values of the transformation (and its derivatives)
+- area below the hat and squeeze function
+- linked-list like reference to the right part of the interval (there will always
+be exactly one interval with right = 0)
+*/
+struct IntervalPoint(S)
```
9il: ~~A user may do not know f1 and f2, but he/she may know their values in required points. The API should support this and be public~~

source/mir/stat/distributions.d
```diff
@@ -0,0 +1,205 @@
+/**
+Random distributions.
```
9il: should be internal for now

source/mir/random/generic/calc.d
```diff
@@ -0,0 +1,157 @@
...
+                a_h += ip.hatA;
+                a_s += ip.squeezeA;
+            }
+        }
+    }
+
+    updateA();
+
+    // Tinflex is not guaranteed to converge
+    for (auto i = 0; i < maxIterations; i++)
```
9il: `foreach` can be used

examples/plot_normal.d
```diff
@@ -0,0 +1,51 @@
...
+import ggplotd.ggplotd;
+import ggplotd.aes;
+import ggplotd.geom;
+
+void main()
+{
+    import std.math : exp, PI, pow, sqrt;
+    import mir.stat.distributions : normalPDF;
+
+    alias π = PI;
```
9il: Please remove any non ASCII chars for name declarations

9il:
Please add `version(LogTinflex)`, which will print report about calculated points

source/mir/random/generic/package.d
joseph-wakeling-sociomantic: > Why not `uniform01`?

There might be some quality issues there, as `uniform01` is a compromise made for cases where speed is prioritized over statistical quality.

Also, won't `uniform(0, 1)` use the integral-type version of `uniform` (meaning this is effectively a coin-toss)?

joseph-wakeling-sociomantic: BTW I would be a bit suspicious about the quality of the implementation of `dice`.  It's a while since I looked in-depth, but I suspect it may be slow.  It might be better to find another way of doing that.

wilzbach:
I adressed most of the points @9il pointed out and I will adress the remaining tomorrow (see list below). Apart from the renaming the three biggest changes are:
- InflectionPoints are processed now as doubly linked list
- Limited all algorithms to floating point only (there's no point in using this algorithm with integers)
- `Infinite` points are supported (beta)

As the newest change contains the renaming to `tinflex` (and moved everything into `internal` and thus hides all the comments, here are the asked question:

## Questions

> @wilzbach Yes. Please share your normal distribution test and charts to see where it is wrong.

As you already realized it's, the test is included in the PR and the charts are essentially rubish :S

image

> Also I am interesting how use should choose interval points. Do more point give better random numbers?

Afaik no. Tinflex is able to generate more intervals based on the starting intervals. For the example here, the algorithm generates about 50 intervals.

> are it correct points?

As you mentioned, ideally we should use `[-real.infinity, 0, real.infinity]`, but the problem here is that it's derivates will return `NaN` for `infinity`. 
Btw as this is a bit related, I used Wolfram Alpha to ensure I don't make a mistake for the derivation of the PDF of the normal distribution:

First derivative: http://www.wolframalpha.com/input/?i=D%5B1%2F(E%5E((x+-+%CE%BC)%5E2%2F(2+%CF%83%5E2))+Sqrt%5B2+Pi%5D+%CF%83),+x%5D

Second derivative: http://www.wolframalpha.com/input/?i=D%5BD%5B1%2F(E%5E((x+-+%CE%BC)%5E2%2F(2+%CF%83%5E2))+Sqrt%5B2+Pi%5D+%CF%83),+x%5D,+x%5D

> tinflex function should accept already calculated IntervalPoint too

As mentioned above the Tinflex algorithm generates _a lot_ of IntervalPoints given valid starting points.

> can sgn(c) \* (c + 1) / c \* x) be negative?
> If yes, can be (c / (c + 1)) not an integer?

That's directly taken from the paper - I will recheck all formulas again tomorrow.

> Also, won't uniform(0, 1) use the integral-type version of uniform (meaning this is effectively a coin-toss)?

No - it behaves similar to `uniform01` ;-)

# Small Todo-List

## Tomorrow
- [x] link to the T_c description
- [x] explain magic Tinflex constants
- [x] use `copysign` wherever possible
- [x] enable detailed logging with `version(LogTinflex)` (currently I enabled a lot of simple logging)
- [x] check derivates of NormalDistribution (hopefully they are wrong)
- [x] compare speed & statistical quality of `uniform01` vs. `uniform`
- [x] check open/closed intervals for `uniform`
- [x] look into quality of `dice`
- [ ] avoid use `internal` in examples

## At the latest until Sunday
- [x] add more tests for `infinite` inflection points
- [x] recheck formulas from Tinflex paper
- [x] enable more of the tests in `example/plot-generic`

## Monday
- [x] move testing setup to `internal` and run only on a special version `Tinflex_test`

## Tuesday
- [x] finally figure out the problems with the NormalDistributions
- [x] add more advanced distributions to the automatic testing setup

9il:
@wilzbach Use standard normal distribution: this will simplified formulas. ~~And use expansion for derivatives in infinity: http://www.wolframalpha.com/input/?i=lim+D%5BD%5B1%2F(E%5E((x)%5E2%2F(2+))+Sqrt%5B2+Pi%5D),+x%5D,+x%5D,+x-%3Einf~~

9il:
derivatives can be just set to 0 starting with `sqrt(FloatType.max).nextDown`

source/mir/random/tinflex/package.d
```diff
@@ -0,0 +1,207 @@
...
+Params:
+    ips = calculated inflection points
+    rng = random number generator to use
+*/
+protected S tfSample(F0, S, RNG)
+          (in F0 f0, in IntervalPoint!S[] ips, in S c, ref RNG rng)
+    if (is(ReturnType!F0 == S) &&
+        isUniformRNG!RNG)
+{
+    import std.algorithm: filter, joiner, map, sum;
```
9il: should not depend on `std.algorithm`

source/mir/random/tinflex/package.d
```diff
@@ -0,0 +1,207 @@
...
+
+    // TODO: filtering not needed anymore
+    auto s = ips[0..$-2].map!`a.hatA`.sum;
```
9il: s does not used

source/mir/random/tinflex/package.d
```diff
@@ -0,0 +1,207 @@
...
+
+        auto hatX = inverse(ips[j].hat(X), c);
+        double squeezeX = ips[j].squeezeA > 0 ? inverse(ips[j].squeeze(X), c) : 0;
+
+        immutable t = u * hatX;
+        if (t <= squeezeX)
+        {
+            break;
+        }
+        else if (t <= exp(f0(X)))
```
9il: do not need `else` here

source/mir/random/tinflex/package.d
```diff
@@ -0,0 +1,207 @@
...
+            X = ips[j].hat._y + (inverseAntiderivative(antiderivative(ips[j].hat(ips[j].x), c)
+                          + ips[j].hat.slope * u, c) - ips[j].hat.a) / ips[j].hat.slope;
+        }
+        else
+        {
+            X = (1-u) * ips[j].x + u * ips[j + 1].x;
+        }
+
+        auto hatX = inverse(ips[j].hat(X), c);
+        double squeezeX = ips[j].squeezeA > 0 ? inverse(ips[j].squeeze(X), c) : 0;
```
9il: Why `double`?

source/mir/random/tinflex/package.d
```diff
@@ -0,0 +1,207 @@
...
+    double X;
+    // acceptance-rejection sampling
+    for (;;)
+    {
+        import std.stdio;
+        writeln(areas);
+        auto j = dice(rng, areas);
+        auto u = uniform01(rng);
+
+        if (abs(ips[j].hat.slope) > 1e-10)
```
9il: This constant should depend on FP type, thought

source/mir/random/tinflex/package.d
```diff
@@ -0,0 +1,207 @@
...
+
+    import mir.random.tinflex.internal.transformations : inverse, antiderivative, inverseAntiderivative;
+
+    double X;
+    // acceptance-rejection sampling
+    for (;;)
+    {
+        import std.stdio;
+        writeln(areas);
+        auto j = dice(rng, areas);
```
9il: `dice` should not be used here. The complexity of `dice` equal to `O(areas.length)`. Please find a way with complexity equal to `O(1)`

source/mir/random/tinflex/package.d
```diff
@@ -0,0 +1,207 @@
...
+        auto u = uniform01(rng);
```
9il: should be `uniform01!S(rng);`

source/mir/random/tinflex/package.d
9il: ok, maybe `O(1)` is something impossible, but using precomputed cumulative sums of normaized lengths, `uniform01` and binary search `O(log(n))` is possible

source/mir/random/tinflex/package.d
```diff
@@ -0,0 +1,207 @@
...
+    // density function
+    private const F0 _f0;
+
+    // generated partition points
+    private const IntervalPoint!S[] _ips;
+
+    // global T_c family
+    private const S c;
+
+    private this(const F0 f0, const IntervalPoint!S[] ips, const S c)
```
9il: make this `protected`

source/mir/random/tinflex/package.d
```diff
@@ -0,0 +1,207 @@
...
+
+/**
+Sample from the distribution.
+Params:
+    ips = calculated inflection points
+    rng = random number generator to use
+*/
+protected S tfSample(F0, S, RNG)
+          (in F0 f0, in IntervalPoint!S[] ips, in S c, ref RNG rng)
+    if (is(ReturnType!F0 == S) &&
```
9il: 1. This constraint should be moved to the struct declaration.
2. `is(ReturnType!F0 == S)`: FP may return any FP type, not only S.

source/mir/random/tinflex/package.d
```diff
@@ -0,0 +1,207 @@
...
+    }
+}
+
+/**
+Sample from the distribution.
+Params:
+    ips = calculated inflection points
+    rng = random number generator to use
+*/
+protected S tfSample(F0, S, RNG)
```
9il: `tinflexImpl`

source/mir/random/tinflex/package.d
```diff
@@ -0,0 +1,207 @@
...
+        auto u = uniform01(rng);
+
+        if (abs(ips[j].hat.slope) > 1e-10)
+        {
+            X = ips[j].hat._y + (inverseAntiderivative(antiderivative(ips[j].hat(ips[j].x), c)
+                          + ips[j].hat.slope * u, c) - ips[j].hat.a) / ips[j].hat.slope;
+        }
+        else
+        {
+            X = (1-u) * ips[j].x + u * ips[j + 1].x;
```
9il: is `ips[j + 1]` is always legal here?

source/mir/random/tinflex/package.d
9il: binary search is part of `std.range.SortedRange`

source/mir/random/tinflex/package.d
```diff
@@ -0,0 +1,207 @@
...
+    auto s = ips[0..$-2].map!`a.hatA`.sum;
+    auto areas = ips[0..$-2].map!((x) => x.hatA / s);
+
+    import std.random: dice, uniform01;
+    import std.math: abs;
+    import mir.internal.math: exp;
+
+    import mir.random.tinflex.internal.transformations : inverse, antiderivative, inverseAntiderivative;
+
+    double X;
```
9il: `S x = void;`

source/mir/random/tinflex/internal/transformations.d
```diff
@@ -0,0 +1,147 @@
...
+
+/**
+Compute inverse transformation of a T_c family given point x.
+From: Table 1, column 3
+*/
+S inverse(S)(in S x, in S c)
+{
+    import mir.internal.math : exp;
+    import std.math : sgn;
+    if (c == 0)
```
9il, line 108: Replace it with `switch`

source/mir/random/tinflex/internal/transformations.d
```diff
@@ -0,0 +1,147 @@
...
+        return exp(x);
+    else if (c == -0.5)
+        return 1 / x ^^ x;
+    else if (c == 1)
+        return x;
+    else
+        return (sgn(c) * x) ^^ (1 / c);
```
9il: This equal to `return pow(x, 1 / fabs(c));`

9il: avoid any use of `^^` please

source/mir/random/tinflex/internal/transformations.d
```diff
@@ -0,0 +1,147 @@
...
+
+/**
+Compute inverse transformation of antiderivative T_c family given point x.
+Table 1, column 5
+*/
+S inverseAntiderivative(S)(in S x, in S c)
+{
+    import mir.internal.math : exp, log;
+    import std.math : sgn;
+    if (c == 0)
```
9il, line 144: `switch` please

source/mir/random/tinflex/internal/transformations.d
```diff
@@ -0,0 +1,147 @@
...
+        return log(x);
+    else if (c == -0.5)
+        return -1 / x;
+    else if (c == -1)
+        return exp(x);
+    else
+        return sgn(c) * (sgn(c) * (c + 1) / c * x) ^^ (c / (c + 1));
```
9il: this is equal to 

``` d
auto d = c + 1;
copysign(pow(d / fabs(c) * x, c / d), c);
```

source/mir/random/tinflex/package.d
dnadlinger: Why not?

source/mir/random/tinflex/package.d
dnadlinger: What does `protected` even mean on a struct?

source/mir/random/tinflex/internal/calc.d
```diff
@@ -0,0 +1,292 @@
...
+    import mir.random.tinflex.internal.transformations : transformToInterval;
+    import std.container.dlist : DList;
+    import std.range : dropBackOne;
+
+    auto intervalTransform = transformToInterval(f0, f1, f2, c);
+
+    auto ips = DList!(IntervalPoint!S)();
+    foreach (i, p; points)
+    {
+        import std.algorithm.mutation : move;
```
9il: note used

source/mir/random/tinflex/internal/calc.d
```diff
@@ -0,0 +1,292 @@
...
+        ips.insertBack(iv);
+    }
+
+    auto nrIntervals = points.length;
+
+    import std.stdio;
+
+    S a_h;
+    S a_s;
+    void updateA()
```
9il: it can be updated on the fly

source/mir/random/tinflex/internal/calc.d
```diff
@@ -0,0 +1,292 @@
...
+    f1 = first derivative of f0
+    f1 = second derivative of f0
+    c = T_c family
+    points = non-overlapping partitioning with at most one inflection point per interval
+    rho = efficiency of the Tinflex algorithm
+    maxIterations = maximal number of iterations before Tinflex is aborted
+
+Returns: Array of IntervalPoints
+*/
+protected IntervalPoint!S[] calcPoints(F0, F1, F2, S)
```
9il: Again, bad complexity

source/mir/random/tinflex/internal/transformations.d
```diff
@@ -0,0 +1,147 @@
+module mir.random.tinflex.internal.transformations;
```
9il, line 1: Please rewrite this module using ~~`switch`~~ `if` only (if `return` is used), `pow`, and `copysign`. It contains computation errors!

source/mir/random/tinflex/package.d
wilzbach: maybe bad naming. renamed it to `totalAreaSum`.
line below:

```
auto areas = ips[0..$-2].map!((x) => x.hatA / s);
```

source/mir/random/tinflex/internal/types.d
```diff
@@ -0,0 +1,318 @@
...
+    /// left position of the interval
+    immutable S x;
+
+    /// T_c family of the interval
+    immutable S c;
+
+    /// transformed values
+    immutable S tx, t1x, t2x;
+
+    this (in S tx, in S t1x, in S t2x, in S x, in S c)
```
9il: Constructor does not required (it has just reordering)

source/mir/random/tinflex/internal/types.d
```diff
@@ -0,0 +1,318 @@
...
+        this.tx = tx;
+        this.t1x = t1x;
+        this.t2x = t2x;
+        this.x = x;
+        this.c = c;
+        import std.stdio;
+        writeln("CC", c);
+    }
+
+    FunType type;
```
9il: should be the last in varaible declarations

source/mir/random/tinflex/internal/types.d
```diff
@@ -0,0 +1,318 @@
...
+                assert(!v.isNaN, "variable " ~ seq[i].stringof ~ " isn't allowed to be NaN");
+            }
+        }
+    }
+}
+
+/// ditto
+IntervalPoint!S intervalPoint(F0, F1, F2, S)
+                             (in F0 f0, in F1 f1, in F2 f2, in S x, in S c = 1.5)
+    if (is(ReturnType!F0 == S) && is(ReturnType!F1 == S) && is(ReturnType!F2 == S) &&
```
9il: remove constraints from simple internal functions

source/mir/random/tinflex/internal/types.d
```diff
@@ -0,0 +1,318 @@
...
+            foreach (i, v; seq)
+            {
+                assert(!v.isNaN, "variable " ~ seq[i].stringof ~ " isn't allowed to be NaN");
+            }
+        }
+    }
+}
+
+/// ditto
+IntervalPoint!S intervalPoint(F0, F1, F2, S)
```
9il: just remove this prototype

source/mir/random/tinflex/internal/types.d
```diff
@@ -0,0 +1,318 @@
...
+                             (in F0 f0, in F1 f1, in F2 f2, in S x, in S c = 1.5)
+    if (is(ReturnType!F0 == S) && is(ReturnType!F1 == S) && is(ReturnType!F2 == S) &&
+        (isFloatingPoint!S))
+{
+    return intervalPoint(f0(x), f1(x), f2(x), x, c);
+}
+
+IntervalPoint!S intervalPoint(S)
```
9il: and this prototype too (remove)

source/mir/random/tinflex/internal/types.d
```diff
@@ -0,0 +1,318 @@
...
+*/
+enum FunType {T1a, T1b, T2a, T2b, T3a, T3b, T4a, T4b}
+
+/**
+Checks whether the function type is concave
+*/
+bool isConcave(in FunType t) pure @safe @nogc nothrow
+{
+    with(FunType)
+    return t == T1a || t == T2a || t == T3a || t == T4a;
```
9il: you forgot to optimise this with my suggestion. please find it in comment for previous commits

source/mir/random/tinflex/internal/types.d
```diff
@@ -0,0 +1,318 @@
...
+- a: concAve
+- b: convex
+- Type 4 is the pure case without any inflection point
+*/
+enum FunType {T1a, T1b, T2a, T2b, T3a, T3b, T4a, T4b}
+
+/**
+Checks whether the function type is concave
+*/
+bool isConcave(in FunType t) pure @safe @nogc nothrow
```
9il: this is not used

source/mir/random/tinflex/internal/types.d
```diff
@@ -0,0 +1,318 @@
...
+{
+    with(FunType)
+    return t == T1a || t == T2a || t == T3a || t == T4a;
+}
+
+/**
+Checks whether the function type is convex
+*/
+bool isConvex(FunType t) pure @safe @nogc nothrow
```
9il: this is not used

source/mir/random/tinflex/internal/types.d
```diff
@@ -0,0 +1,318 @@
...
+    return determineType(intervalPoint(f0, f1, f2, bl), intervalPoint(f0, f1, f2, br));
+}
+
+/// ditto
+FunType determineType(S)(in IntervalPoint!S l, in IntervalPoint!S r)
+    if (isFloatingPoint!S)
+{
+    import std.traits : isFloatingPoint;
+
+    static if (isFloatingPoint!S)
```
9il: please review your code and remove rubish

source/mir/random/tinflex/package.d
9il: oh, yes. Should be just `package`

source/mir/random/tinflex/package.d
9il: By proper design this function should not have any code that can use something from `std.algorithm`. Is needs only binary search from `std.range.SortedRange`.

source/mir/random/tinflex/internal/types.d
```diff
@@ -0,0 +1,318 @@
...
+
+    LinearFun!S hat;
+    LinearFun!S squeeze;
+
+    // make NaN
+    S hatA = 0;
+    S squeezeA = 0;
+
+    // disallow NaN points
+    invariant {
```
9il, line 54: when this invariant is checked?

source/mir/random/tinflex/internal/types.d
```diff
@@ -0,0 +1,318 @@
...
+}
+
+/**
+Determine the function type of an interval.
+Based on Theorem 1 of the Tinflex paper.
+Params:
+    bl = left side of the interval
+    br = right side of the interval
+*/
+FunType determineType(F0, F1, F2, S)
```
9il: Remove this prototype

source/mir/random/tinflex/internal/types.d
```diff
@@ -0,0 +1,318 @@
...
+            (l.c <= 0 && isInfinity(r.tx) && r.tx < 0 ))
+        {
+            writeln("infC");
+            writeln("L", l);
+            writeln("R", r);
+            if (r.t2x < 0 && r.t1x > 0)
+                return T4a;
+            if (r.t2x > 0 && r.t1x > 0)
+                return T4b;
+            assert(0, "Invalid case");
```
9il: why not simple code like : 

``` d
assert(r.t1x > 0, "some informative text");
assert(r.t2x, "some informative text");
return r.t2x < 0 ? T4a : T4b;
```

?

source/mir/random/tinflex/internal/types.d
```diff
@@ -0,0 +1,318 @@
...
+                return T3a;
+            if (l.t2x > 0 && r.t2x < 0)
+                return T3b;
+        }
+        if (l.t2x <= 0 && r.t2x <= 0)
+            return T4a;
+        if (l.t2x >= 0 && r.t2x >= 0)
+            return T4b;
+    }
+    assert(0, "Unknown type");
```
9il: ditto, why you need this in the end? Can it be rewrited without `assert(0, )`?

source/mir/random/tinflex/internal/types.d
```diff
@@ -0,0 +1,318 @@
...
+                     (in F0 f0, in F1 f1, in F2 f2, in S bl, in S br)
+    if (is(ReturnType!F0 == S) && is(ReturnType!F1 == S) && is(ReturnType!F2 == S) &&
+        (isFloatingPoint!S))
+{
+    return determineType(intervalPoint(f0, f1, f2, bl), intervalPoint(f0, f1, f2, br));
+}
+
+/// ditto
+FunType determineType(S)(in IntervalPoint!S l, in IntervalPoint!S r)
+    if (isFloatingPoint!S)
```
9il: please rewrite it with proper `assert`ions

source/mir/random/tinflex/internal/types.d
```diff
@@ -0,0 +1,318 @@
...
+FunType determineType(F0, F1, F2, S)
+                     (in F0 f0, in F1 f1, in F2 f2, in S bl, in S br)
+    if (is(ReturnType!F0 == S) && is(ReturnType!F1 == S) && is(ReturnType!F2 == S) &&
+        (isFloatingPoint!S))
+{
+    return determineType(intervalPoint(f0, f1, f2, bl), intervalPoint(f0, f1, f2, br));
+}
+
+/// ditto
+FunType determineType(S)(in IntervalPoint!S l, in IntervalPoint!S r)
```
9il: This function does not follow `Theorem 2`. Please rewrite it with proper way

source/mir/random/tinflex/internal/transformations.d
wilzbach: are you sure?
It is taken from this table and a combination of two cases - hence the sgn in the power matters?

image

edit: the problem is that as soon as c is negative, exponenation would return a complex number and thus yields `NaN`.

9il: fix `c=%.3f"`

9il: just `copysign(pow(d / c * x, c / d), c);` works

source/mir/random/tinflex/internal/types.d
wilzbach, line 54: after the constructor

source/mir/random/tinflex/internal/types.d
wilzbach: > Constructor does not required (it has just reordering)

My motivation was that the invariant check (against NaNs) is called after the constructor.

source/mir/random/tinflex/internal/linearfun.d
```diff
@@ -0,0 +1,125 @@
...
+unittest
+{
+    assert(secant((double x) => x ^^ 2, 1., 3) == linearFun(4., -3));
+    assert(secant((double x) => x ^^ 2, 3., 5) == linearFun(8., -15));
+}
+
+/**
+Calculate tangent of any point (x, y) given it's derivate f1
+*/
+LinearFun!(typeof(F1.init(S.init))) tangent(F1, S)(in F1 f1, in S x, in S y)
```
9il: Should return  `LinearFun! S`

wilzbach: > Should return LinearFun! S

F1 might return a different type than S.

source/mir/random/tinflex/internal/types.d
```diff
@@ -0,0 +1,318 @@
...
+        import std.math : isInfinity, isNaN;
+    }
+    else
+    {
+        alias isInfinity = (x) => false;
+        alias isNaN = (x) => false;
+    }
+
+
+    assert(l.x < r.x, "invalid interval");
```
9il: `<=`

source/mir/random/tinflex/internal/types.d
```diff
@@ -0,0 +1,318 @@
...
+        if (isInfinity(r.tx))
+        {
+            writeln("infB");
+            if (l.t2x < 0 && l.tx > 0)
+                return T4a;
+            else
+                assert(0, "Invalid case");
+        }
+
+        if ((l.c > 0 && r.tx == 0.0) ||
```
9il: Do not need to handle +-infinity here. +- infinity should be replaced with +-S.max before the iteration

source/mir/random/tinflex/package.d
```diff
@@ -142,31 +144,30 @@ Params:
...
+    auto areas = ips[0..$-2].map!((x) => x.hatA / totalAreaSum);
 
     import std.random: dice, uniform01;
     import std.math: abs;
     import mir.internal.math: exp;
 
     import mir.random.tinflex.internal.transformations : inverse, antiderivative, inverseAntiderivative;
 
-    double X;
+    double X = void;
```
9il: Why `double`

source/mir/random/tinflex/package.d
```diff
@@ -142,31 +144,30 @@ Params:
     ips = calculated inflection points
     rng = random number generator to use
 */
-protected S tfSample(F0, S, RNG)
+protected S tinflexImpl(F0, S, RNG)
```
9il: private

source/mir/random/tinflex/internal/types.d
```diff
@@ -205,6 +159,16 @@ FunType determineType(S)(in IntervalPoint!S l, in IntervalPoint!S r)
     assert(0, "Unknown type");
 }
 
+/// convenience wrapper for unittests
+version(unittest) FunType determineType(F0, F1, F2, S)
```
9il: remove

source/mir/random/tinflex/internal/transformations.d
```diff
@@ -99,16 +110,15 @@ From: Table 1, column 3
 */
 S inverse(S)(in S x, in S c)
 {
-    import mir.internal.math : exp;
-    import std.math : sgn;
+    import mir.internal.math : exp, pow;
+    import std.math : fabs;
```
9il: `mir.internal.math` has not `fabs` ?

source/mir/random/tinflex/internal/transformations.d
```diff
@@ -69,16 +77,19 @@ Table 1, column 5
...
     else if (c == -0.5)
         return -1 / x;
     else if (c == -1)
         return -log(-x);
+    auto d = c + 1;
+    if (c > 0)
+        return c / d * pow(x, d / c);
     else
-        return sgn(c) * c / (c + 1) * (sgn(c) * x)^^ ((c + 1) / c);
+        return - c / d * pow(-x, d / c);
```
9il: This can be simplified with as `fabs(c / d) * pow(x, d / c);`, pow will ignore sign

source/mir/random/tinflex/internal/area.d
```diff
@@ -0,0 +1,288 @@
...
+    // sh.y is the boundary point where f obtains its maximum
+
+    // specializations for T_c family (page 6)
+    if (c == 0)
+    {
+        // T_c = log(x)
+        // Error in table, see equation (4)
+        immutable z = s * sh.slope * (r - l);
+        // check whether approximation is possible, page 5
+        if (abs(z) < 1e-6)
```
9il: Why `1e-6`?

source/mir/random/tinflex/internal/transformations.d
```diff
@@ -0,0 +1,157 @@
...
+        IntervalPoint!S opCall(S x)
+        {
+            return IntervalPoint!S(t0(x), t1(x), t2(x), x, c);
+        }
+        auto t0 (S)(S x) const
+        {
+            if (c == 0)
+                return f0(x);
+            else
+                return sgn(c) * exp(c * f0(x));
```
9il: copysign

source/mir/random/tinflex/internal/area.d
```diff
@@ -0,0 +1,288 @@
...
+            // F_T = e^x
+            area = (exp(sh(r)) - exp(sh(l))) / sh.slope;
+        }
+    }
+    else
+    {
+        // prevent numeric errors
+        if (sgn(c) * sh(r) < 0 || sgn(c) * sh(l) < 0)
+        {
+            assert("shouldn't happen");
```
9il: ???

source/mir/random/tinflex/internal/transformations.d
```diff
@@ -0,0 +1,157 @@
...
+        return exp(x);
+    else if (c == -0.5)
+        return -1 / x;
+    else if (c == -1)
+        return -log(-x);
+    auto d = c + 1;
+    if (c > 0)
+        return c / d * pow(x, d / c);
+    else
+        return - c / d * pow(-x, d / c);
```
9il: `return fabs(c) / d * pow(fabs(x), d / c);`

source/mir/random/tinflex/internal/transformations.d
```diff
@@ -0,0 +1,157 @@
...
+    import mir.internal.math : exp, log, pow;
+    import std.math : copysign, fabs;
+    if (c == 0)
+        return log(x);
+    else if (c == -0.5)
+        return -1 / x;
+    else if (c == -1)
+        return exp(x);
+    immutable d = c + 1;
+    return copysign(pow(d / fabs(c) * x, c / d), c);
```
9il: after reviewing R sources: should be `copysign(pow(fabs(d / c * x) , c / d), c)`

9il: pow may have different implementation for negative numbers :-/

source/mir/random/tinflex/internal/calc.d
```diff
@@ -0,0 +1,190 @@
...
+    foreach (i; 0..maxIterations)
+    {
+        // Tinflex aims for a user defined efficiency
+        if (totalHatArea / totalSqueezeArea <= rho)
+            break;
+
+        S a_avg = (totalHatArea - totalSqueezeArea) / (nrIntervals - 1);
+        // first iteration: search only (we update the list online later)
+        auto it = ips[];
+        foreach (j; 0..nrIntervals - 1)
```
9il: Do not use range primitives. They hide bugs here

source/mir/random/tinflex/internal/calc.d
```diff
@@ -0,0 +1,190 @@
...
+        {
+            if (it.front.hatArea - it.front.squeezeArea > a_avg)
+            {
+                // prepare total areas for update
+                totalHatArea -= it.front.hatArea;
+                totalSqueezeArea -= it.front.squeezeArea;
+
+                auto nextView = it.save.dropOne;
```
9il:
``` d
auto itSave = it;
it.popFront; // do not move this to the end of the loop!
auto left = itSave.front;
auto right = it.front;
```

source/mir/random/tinflex/internal/calc.d
```diff
@@ -0,0 +1,190 @@
...
+                itNext.popFront();
+                ips.insertBefore(itNext, midIP);
+
+                // update total areas
+                totalHatArea += it.front.hatArea + midIP.hatArea;
+                totalSqueezeArea += it.front.squeezeArea + midIP.squeezeArea;
+
+                nrIntervals++;
+            }
+            it.popFront();
```
9il: remove

source/mir/random/tinflex/internal/calc.d
```diff
@@ -0,0 +1,190 @@
...
+                calcInterval(it.front, midIP);
+
+                // insert new middle part into linked list
+                auto itNext = it.save;
+                itNext.popFront();
+                ips.insertBefore(itNext, midIP);
+
+                // update total areas
+                totalHatArea += it.front.hatArea + midIP.hatArea;
+                totalSqueezeArea += it.front.squeezeArea + midIP.squeezeArea;
```
9il:
``` d
                // split the interval at the arcmean into two parts
                auto mid = arcmean(left.x, right.x);
                IntervalPoint!S midIP = intervalTransform(mid, left.c);

                // recalculate intervals
                calcInterval(midIP, *right);
                calcInterval(*left, midIP);

                // insert new middle part into linked list
                ips.insertAfter(itSave, midIP);

                // update total areas
                totalHatArea += left.hatArea + midIP.hatArea;
                totalSqueezeArea += left.squeezeArea + midIP.squeezeArea;
```

source/mir/random/tinflex/internal/area.d
```diff
@@ -226,9 +226,11 @@ body
             }
         }
     }
-    import std.math : isInfinity, isNaN;
-    if (isInfinity(area) || isNaN(area))
+    import std.math : isNaN;
+    if (isNaN(area))
```
9il: see R code:

``` r
area <- function(cT, a,b,y, from,to) {
  ## -----------------------------------------------------------------------
  ## Compute area underneath hat or squeeze in particular interval.
  ## ------------------------------------------------------------------------
  ##   cT       ... parameter for transformation
  ##   a, b, y  ... intercept, slope and anchor point of transformed hat or squeeze
  ##   from, to ... range of integration
  ## ------------------------------------------------------------------------
  ## Return: area;
  ##         0 in case of an error.
  ## ------------------------------------------------------------------------

  ## Compute area ...
  area <- do.area(cT,a,b,y,from,to)

  ## ... and check result.

  if (! is.finite(area))
    ## We return Inf in all cases where 'area' is not finite (e.g. NaN or NA)
    area <- Inf

  if (area < 0) {
    ## Area is strictly negative.
    ## Two possible reasons:
    ## 1. The correct value of 'area' is extremely small or 0.
    ##    The negative number might result from cancelation.
    ##    Then 'area' can be set to 0.
    ## 2. Due to sever round-off error, the result is numerical garbage.
    ##
    ## We observed (2) in our experiments. So we discard the result and
    ## return 'Inf' (which means that the interval must be split).

    area <- Inf
  }

  return (area)
}
```

9il:
``` d
    import std.math : isFinite;
    if (!isFinite(area))
        area = S.infinity;
    else if (area < 0)
        area = S.infinity;
    return area;
```

wilzbach:
@joseph-wakeling-sociomantic @WebDrake I pushed another PR as it got quite messy here.
Please see #240 for the continuation.

PR #225, opened 2016-06-09
fix alias in inernal.math

9il:


source/mir/internal/math.d
```diff
@@ -14,7 +14,7 @@ version(LDC)
 
     alias powi = llvm_powi;
 
-    alias pow = llvm_pow;
+    alias pow = llvm_powi;
 
     alias pow = llvm_pow;
```
wilzbach, line 19: dumb question: why do we have the alias `pow` twice?

9il, line 19: one for fp version, one for integer version

wilzbach:
> one for fp version, one for integer version

Alrighty - thanks for the explanation!

Nitpick: There is a typo in your commit message, you want to fix it?
Feel free to directly merge afterwards ;-)

PR #226, opened 2016-06-10
Add dub header to examples

wilzbach:
Adds the new dub header file to the examples and allows the examples to be directly executed like.
- At some point hopefully specifying a name will be optional
- Should we use `>0.15` instead? (See: https://github.com/dlang/dub/wiki/Version-management)

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -1,7 +1,7 @@
 machine:
   environment:
     DMD: 2.071.0
-    DUB: 0.9.25
+    DUB: 1.0.0-beta.1
     PATH: "${HOME}/dmd2/linux/bin64:${PATH}"
     LD_LIBRARY_PATH: "${HOME}/dmd2/linux/lib64:${LD_LIBRARY_PATH}"
 checkout:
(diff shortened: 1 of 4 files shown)


9il:
> Should we use >0.15 instead? (See: https://github.com/dlang/dub/wiki/Version-management)

Yes

wilzbach:
> Yes

Alrighty - rebased ;-)

9il:
have you test it?

wilzbach:
> have you test it?

Let's let automate that!
I added it to the CircleCi setup.
(N.B.: It only tests whether compilation is successful - it doesn't run them.)

wilzbach:
> Let's let automate that!
> I added it to the CircleCi setup.

Damn - I forgot that it's still part of the beta dub release.
Bumped the dub version for CircleCi ;-)

> File source/mir/sum.d (not in diff) was modified

I have no idea about this message, though :/

wilzbach:
all tests are passing and the bug in CodeCov is gone :)

PR #227, opened 2016-06-10
Make Hoffman example runnable and add to CI

wilzbach:
I just realized that currently the hoffman example doesn't compile, we probably should add it to the CI setup ;-)
While I was at it, I modified the file, s.t. it can take an input directory or alternative use the executable folder to find the data dir.


9il:
Thanks!

Issue #120, opened 2016-04-16
mir.las.sum: assertion failure on Win64 with ldc


wilzbach:
It yields:

```
core.exception.AssertError@source\mir\las\sum.d(1835): Assertion failure
```

[line](https://github.com/DlangScience/mir/blob/master/source/mir/las/sum.d#L1835)
[details](https://ci.appveyor.com/project/9il/mir/build/1.0.100/job/j4sav10gb62kuhf7)

See also: #116 
##

9il:
This looks like LDC bug

wilzbach:
Ok then I will add `allow_failures` temporarily to the appveyor script :(
Do you submit it to ldc?

9il:
> Do you submit it to ldc?

https://github.com/ldc-developers/ldc/issues/1436

PR #228, opened 2016-06-14
Update appveyor.yml to use latest LDC and DUB

PetarKirov:


PetarKirov:
I was reading http://wiki.dlang.org/Continuous_Integration (btw, nice work @wilzbach) and I stumbled upon: `...Should be removed once dub 0.9.26 is available`, so I wanted to test if the problem was resolved. I hope you don't mind :)

wilzbach:
> so I wanted to test if the problem was resolved. I hope you don't mind

For mir we have another issue with ldc on windows - see #120 :/

PetarKirov:
> For mir we have another issue with ldc on windows - see #120 :/

That's quite unfortunate, though it looks like the issue is not is not visible on [AppVeyor](https://ci.appveyor.com/project/9il/mir/build/1.0.105).

Should I leave `allow_failures` on for now?

wilzbach:
> That's quite unfortunate, though it looks like the issue is not is not visible on AppVeyor.

Then I guess the bug got accidentally fixed. That's great news!

> Should I leave allow_failures on for now?

No no, better update the [compatibility table](https://github.com/libmir/mir#compatibility) later :)

@9il - let's go ahead & merge?

PetarKirov:
It seems that the LDC [bug](https://github.com/ldc-developers/ldc/issues/1436) is still present, just encapsulated in a [try...catch](https://github.com/libmir/mir/blob/5067cc3b848170a79585dc737d8c8cb0349dbf95/source/mir/sum.d#L1865). You can see this on the output of the builds. The issue is only present with [LDC](https://ci.appveyor.com/project/9il/mir/build/1.0.105/job/avy1xu20cl8httjw#L63) (both 0.17 and 1.0.0).

Since @9il worked around the issue, I think it's better to leave the checks on, so LDC can get more extensive testing.

PetarKirov:
> No no, better update the compatibility table

@wilzbach see PR https://github.com/libmir/mir/pull/229.

9il:
> It seems that the LDC bug is still present, just encapsulated in a try...catch. You can see this on the output of the builds. The issue is only present with LDC (both 0.17 and 1.0.0).

Thanks! Probably forgot to remove this block

9il:
see #230

PetarKirov:
@9il so should I keep `allow_failures` for LDC or not?

9il:
> @9il so should I keep allow_failures for LDC or not?

Yes, keep please

PetarKirov:
@9il Alright, fixed & rebased.

PetarKirov:
Thanks

PR #231, opened 2016-06-15
Update iteration.d

trikko:
Typo


wilzbach:
Good catch - thanks :)

trikko:
Too bad. It was the first result on google for keyword "descriprottion"...

PR #234, opened 2016-06-25
Add std.random.discrete: Discrete sampler

### LINKED ISSUE #233: mir.random.discrete
- [x] `naiveDiscrete`
- [x] `discrete` (using alias method) #237 
- [ ] Bernoulli
- [ ] binomial
- [ ] negative binomial
- [ ] geometric
- [ ] hypergeometric
- [ ] Poisson
##

wilzbach:
[preview docs](http://preview.mir.rocks/mir_random_discrete.html)


source/mir/random/mixture.d
```diff
@@ -0,0 +1,112 @@
...
+    {
+        import std.random : rndGen;
+        return opCall(rndGen);
+    }
+
+    /// samples a value from the discrete distribution using a custom random generator
+    size_t opCall(RNG)(ref RNG gen) const
+    {
+        import std.random : uniform;
+        T v = uniform!"[]"(0, _cdPoints[$-1], gen);
```
9il: it can be just `uniform`

9il: without `"[]"`

source/mir/random/mixture.d
```diff
@@ -0,0 +1,112 @@
...
+        return (cast(SortedRange!(const(T)[], "a < b")) r).lowerBound(v).length;
+    }
+}
+
+/// ditto
```
9il: This should be above the struct

source/mir/random/mixture.d
```diff
@@ -0,0 +1,112 @@
...
+Discrete!T discrete(T)(const(T)[] cdPoints)
+{
+    return Discrete!T(cdPoints);
+}
+
+///
+unittest
```
9il: ... along with the unittest

source/mir/random/mixture.d
```diff
@@ -0,0 +1,112 @@
...
+    // sample from the discrete distribution
+    uint[size_t] obs;
+    foreach (i; 0..10_000)
+        obs[ds()]++;
+}
+
+unittest
+{
+    import std.random : rndGen;
+    rndGen.seed(42);
```
9il: Should be used named RNG because `rndGen` may be changed in the future

source/mir/random/mixture.d
```diff
@@ -0,0 +1,112 @@
...
+    {
+        import std.random : rndGen;
+        return opCall(rndGen);
+    }
+
+    /// samples a value from the discrete distribution using a custom random generator
+    size_t opCall(RNG)(ref RNG gen) const
+    {
+        import std.random : uniform;
+        T v = uniform(0, _cdPoints[$-1], gen);
```
9il: `uniform!T` please

source/mir/random/mixture.d
```diff
@@ -0,0 +1,112 @@
...
+
+///
+unittest
+{
+    // 10%, 20%, 20%, 40%, 10%
+    auto probs = [0.1, 0.3, 0.5, 0.9, 1];
+    auto ds = discrete(probs);
+
+    // sample from the discrete distribution
+    uint[size_t] obs;
```
9il: auto obs = new uint[probs.length];

source/mir/random/mixture.d
```diff
@@ -0,0 +1,112 @@
...
+unittest
+{
+    import std.random : rndGen;
+    rndGen.seed(42);
+
+    // 10%, 20%, 20%, 40%, 10%
+    auto probs = [0.1, 0.3, 0.5, 0.9, 1];
+    auto ds = discrete(probs);
+
+    uint[size_t] obs;
```
9il: auto obs = new uint[probs.length];

source/mir/random/mixture.d
```diff
@@ -0,0 +1,112 @@
...
+unittest
+{
+    import std.random : rndGen;
+    rndGen.seed(42);
+
+    // 1, 2, 1
+    auto cdPoints = [1, 3, 4];
+    auto ds = discrete(cdPoints);
+
+    uint[size_t] obs;
```
9il: auto obs = new uint[probs.length];

source/mir/random/mixture.d
```diff
@@ -0,0 +1,112 @@
...
+Discrete!T discrete(T)(const(T)[] cdPoints)
+{
+    return Discrete!T(cdPoints);
+}
+
+///
+unittest
+{
+    // 10%, 20%, 20%, 40%, 10%
+    auto probs = [0.1, 0.3, 0.5, 0.9, 1];
```
9il: name probs refers to pdf. You can use `comulativeFold` from std.algorithm

wilzbach: `cumulativeFold` will be published in 2.072

https://github.com/dlang/phobos/pull/3972

source/mir/random/mixture.d
```diff
@@ -0,0 +1,112 @@
...
+    auto ds = discrete(probs);
+
+    // sample from the discrete distribution
+    uint[size_t] obs;
+    foreach (i; 0..10_000)
```
9il: 1000 is enought

source/mir/random/mixture.d
wilzbach: doesn't work - the prototype is `auto uniform(string boundaries = "[)", T1, T2)(T1 a, T2 b)`

https://dlang.org/phobos/std_random.html#.uniform

9il:
ping @joseph-wakeling-sociomantic

source/mir/random/mixture.d
9il: ok, then just another name

source/mir/random/mixture.d
wilzbach: ?

wilzbach:
Docs can be previewed at [preview.mir.rocks](http://preview.mir.rocks/mir_random_discrete.html).

> ping @joseph-wakeling-sociomantic

@WebDrake you told me to ping both accounts?

source/mir/random/mixture.d
9il: uniform!("[)", T, T) then

source/mir/random/mixture.d
```diff
@@ -0,0 +1,109 @@
...
+using `std.algorithm.cumulativeFold`. Hence `cdPoints` is assumed to be sorted.
+
+Params:
+    cdPoints = cumulative density points
+
+Returns:
+    A sampler that can be called to sample from the distribution.
+*/
+struct Discrete(T)
+    if (isNumeric!T)
```
9il: Struct declaration should be after function. And add a small separate comment please

wilzbach: > Struct declaration should be after function. And add a small separate comment please

ok for the order, but `ditto` would group them:

image

source/mir/random/discrete.d
```diff
@@ -0,0 +1,109 @@
...
+
+Returns:
+    A sampler that can be called to sample from the distribution.
+*/
+Discrete!T discrete(T)(const(T)[] cdPoints)
+{
+    return Discrete!T(cdPoints);
+}
+
+/// ditto
```
9il: Please do not compose function and struct declarations

source/mir/random/discrete.d
```diff
@@ -0,0 +1,109 @@
...
+    auto ds = discrete(cdPoints);
+
+    auto obs = new uint[cdPoints.length];
+    foreach (i; 0..1000)
+        obs[ds(rndGen)]++;
+}
+
+unittest
+{
+    import std.random : Random;
```
9il, line 112: Random alias may change

9il, line 112: Oh, this dose not matter for this tests, nevermind

source/mir/random/discrete.d
```diff
@@ -0,0 +1,123 @@
...
+        import std.random : rndGen;
+        return opCall(rndGen);
+    }
+
+    /// samples a value from the discrete distribution using a custom random generator
+    size_t opCall(RNG)(ref RNG gen) const
+    {
+        import std.random : uniform;
+        T v = uniform!("[)", T, T)(0, _cdPoints[$-1], gen);
+        return (cast(SortedRange!(const(T)[], "a <= b")) r).lowerBound(v).length;
```
WebDrake, line 92: Can't you just use `assumeSorted` here ... ?

9il, line 92: `SortedRange` constructor always check that range is sorted in debug mode. This will affect complexity

WebDrake, line 92: ...  actually, isn't `r` _already_ a sorted range?  Why the `cast` at all?

source/mir/random/discrete.d
```diff
@@ -0,0 +1,123 @@
...
+
+/**
+_Discrete distribution sampler that draws random values from a _discrete
+distribution given an array of the respective cumulative density points.
+`cdPoints` is an array of the cummulative sum over the probabilities
+or counts of a _discrete distribution without a starting zero.
+
+Complexity: O(log n) where n is the number of `cdPoints`.
+*/
+struct Discrete(T)
```
WebDrake, line 52: Did you take at all a look at `hap.random`'s `DiscreteDistribution` when creating this?

source/mir/random/discrete.d
```diff
@@ -0,0 +1,123 @@
...
+    if (isNumeric!T)
+{
+    import std.range : SortedRange;
+
+    private const(T)[] _cdPoints;
+    private SortedRange!(const(T)[], "a <= b") r;
```
WebDrake, line 58: I don't understand the need for the 2 entries here.  `_cdPoints` is barely used and you have all the data you need in `r`.

9il, line 58: @wilzbach fix this please. `SortedRange` should have something like `release`

WebDrake, line 58: BTW, two style notes:
- identifiers with a leading underscore `_likeThis` are IIRC reserved for the either the language or the runtime;
- try to always give a _meaningful_ name to all internal variables, so the person reading the code can understand their purpose.  `r` is a meaningless name and the parameter isn't even commented -- but if you give it a _meaningful_ name, the lack of comment documentation may not matter.

source/mir/random/discrete.d
9il, line 52: no

source/mir/random/discrete.d
```diff
@@ -0,0 +1,123 @@
...
+        import std.random : uniform;
+        T v = uniform!("[)", T, T)(0, _cdPoints[$-1], gen);
+        return (cast(SortedRange!(const(T)[], "a <= b")) r).lowerBound(v).length;
+    }
+}
+
+unittest
+{
+    import std.random : Random;
+    auto rndGen = Random(42);
```
WebDrake, line 99: I would advise not shadowing the name of the standard random generator instance defined in `std.random`, even though you don't import it.  Just call this `gen`.

source/mir/random/discrete.d
```diff
@@ -0,0 +1,123 @@
...
+
+    /**
+    The cumulative density points `cdPoints` are assumed to be sorted and given
+    without a starting zero. They can be calculated with
+    `std.algorithm.iteration.cumulativeFold` from probabilities or counts.
+
+    Params:
+        cdPoints = cumulative density points
+    */
+    this(const(T)[] cdPoints)
```
WebDrake, line 68: Any reason not to give the constructor the relative proportions for the different discrete element selection probabilities (as is done with `dice`) and let the constructor do the work of putting together the sorted range of cumulative values?

I assume one motivation might be to not have the constructor do any memory allocation, but you can get round that by having a second constructor that accepts a `SortedRange`as input (not being too specific about its type; it probably suffices that it's an input range, but is almost certainly safer that it be a forward range).

source/mir/random/discrete.d
```diff
@@ -0,0 +1,123 @@
...
+    {
+        import std.random : rndGen;
+        return opCall(rndGen);
+    }
+
+    /// samples a value from the discrete distribution using a custom random generator
+    size_t opCall(RNG)(ref RNG gen) const
+    {
+        import std.random : uniform;
+        T v = uniform!("[)", T, T)(0, _cdPoints[$-1], gen);
```
WebDrake, line 91: Note that the `.back` property of the `SortedRange` instance would work just as well as `_cdPoints[$-1]`.

source/mir/random/discrete.d
wilzbach, line 92: > Why the cast at all?

because the internal variable `r` has type `const SortedRange!(const(T)[], "a <= b")` - I didn't know another way to get rid of the `const`.

source/mir/random/discrete.d
```diff
@@ -0,0 +1,123 @@
...
+        cdPoints = cumulative density points
+    */
+    this(const(T)[] cdPoints)
+    {
+        _cdPoints = cdPoints;
+        r = typeof(r)(_cdPoints);
+    }
+
+    /// cumulative density points
+    const(T)[] cdPoints() const @property
```
WebDrake, line 75: Is there any need to define this method at all?

source/mir/random/discrete.d
9il, line 68: meaty code here is one line. This looks like bad style to have big API for simple things.

source/mir/random/discrete.d
PetarKirov, line 58: Nothing else, except keywords and version identifiers is reserved by the langauge (furthermore we have a module system, to prevent name conflicts with symbols from druntime). I'm not a fan of `_likeThis` names either, but they're used in many places in phobos for private members. Probably part of the reason was that there were bugs in the module system, though they should all be fixed by now.

Otherwise +1 for meaningful names.

wilzbach, line 58: > @wilzbach fix this please. SortedRange should have something like release

`release` uses `move`, so that won't work? We can just remove `cdPoints` from the public interface?

https://github.com/dlang/phobos/blob/master/std/range/package.d#L8023

wilzbach, line 58: > identifiers with a leading underscore _likeThis are IIRC reserved for the either the language or the runtime;

Isn't that a common style to denote that's a private variables. In this case the leading underscore is used to distinguish between the `private` variable and public getter.

> try to always give a meaningful name to all internal variables, so the person reading the code can understand their purpose. r is a meaningless name and the parameter isn't even commented -- but if you give it a meaningful name, the lack of comment documentation may not matter.

Fair point!

source/mir/random/discrete.d
WebDrake, line 68: > meaty code here is one line. This looks like bad style to have big API for simple things.

I'm not sure I follow what you mean, can you clarify?

source/mir/random/discrete.d
9il, line 75: just more control for user

source/mir/random/discrete.d
WebDrake, line 92: I would repeat my suggestion that you look at `hap.random`'s `DiscreteDistribution` -- I did not need any casts under the hood:
https://github.com/WebDrake/hap/blob/e25fc15a17e4e439333740d2a9894fc7d3c94012/source/hap/random/distribution.d#L323-L329

WebDrake, line 92: BTW I don't see that `r` has type `const SortedRange`, it's not declared `const` ... ?

source/mir/random/discrete.d
WebDrake, line 58: > Nothing else, except keywords and version identifiers is reserved by the langauge (furthermore we have a module system, to prevent name conflicts with symbols from druntime)

I believe it's still considered preferable to use `likeThis_` for library code ... ?

I recognize that leading-underscore names are widely used in phobos, but it's still my understanding that this is technically a style violation.

WebDrake, line 58: Ah, I see the current style guide has: `Also, names do not begin with an underscore ‘_’ unless they are private.`  Ignore that remark, then.

source/mir/random/discrete.d
WebDrake, line 75: Hmmm, if the user needs to care, presumably they know what they provided as input ... ?  I would suggest better to not provide this method unless someone has an explicit use-case.

I also don't think it's a good idea to constrain the choice of cumulative density points to `const(T)[]`.

source/mir/random/discrete.d
9il, line 68: I mean that having huge API for such simple thing like this distribution is too much efforts. API review needs time from a user / reviewer.

source/mir/random/discrete.d
9il, line 92: `opCall` is const

source/mir/random/discrete.d
```diff
@@ -0,0 +1,123 @@
...
+    without a starting zero. They can be calculated with
+    `std.algorithm.iteration.cumulativeFold` from probabilities or counts.
+
+    Params:
+        cdPoints = cumulative density points
+    */
+    this(const(T)[] cdPoints)
+    {
+        _cdPoints = cdPoints;
+        r = typeof(r)(_cdPoints);
```
WebDrake, line 71: This just wraps `_cdPoints`, right -- it doesn't copy it?

Note that the `const`ness of the constructor input provides poor guarantees.  You can't be certain that someone will not subsequently modify the points and mess up the assumptions of the sorted property (it's only the `SortedRange` _constructor_ that checks for sortedness).

source/mir/random/discrete.d
9il, line 75: A user may accept this struct as function argument. Then he want to print intenal stuff for debug without changing the API.

source/mir/random/discrete.d
WebDrake, line 68: Right, but absent any other implementation I would suggest that the `Discrete` API match that for `dice` (its function counterpart), which expects to be given individual proportions, not cumulative ones.

source/mir/random/discrete.d
9il, line 58: > release uses move, so that won't work? We can just remove cdPoints from the public interface?
> 
> https://github.com/dlang/phobos/blob/master/std/range/package.d#L8023

OK, then use `assumeSorted` instead. And make Phobos PR to replace `debug` with `debug(std_range)`. Remember also about to methods that should be static/alias

source/mir/random/discrete.d
wilzbach, line 91: > Note that the .back property of the SortedRange instance would work

... if it would be `const` ?

https://github.com/dlang/phobos/pull/4478

source/mir/random/discrete.d
9il, line 68: Mir random is going to import only basic uniform RNG without any additional bloat like `dice`

9il, line 68: It would also contain replacement for `uniform` function family

source/mir/random/discrete.d
WebDrake, line 92: Right, but that would mean you can't use `lowerBound` with a `const SortedRange` here, right?  Surely that should be fixed in `SortedRange`, rather than covering up the non-`const`-ness with a `cast`?

source/mir/random/discrete.d
WebDrake, line 52: Any reason why not?  There is no licensing issue, and I'm happy for that code to be re-used here.

source/mir/random/discrete.d
wilzbach, line 92: > Surely that should be fixed in SortedRange, rather than covering up the non-const-ness with a cast?

Yes I will try to fix Phobos, but that won't be released before 2.073 :/

source/mir/random/discrete.d
WebDrake, line 68: > Mir random is going to import only basic uniform RNG without any additional bloat like `dice`

Not the point -- designing the API this way it will complicating the use of `Discrete` as a drop-in replacement for existing phobos `dice` (and also throw on the end user the responsibility of summing proportions correctly).

WebDrake, line 68: BTW why replace `std.random.uniform`, what's wrong with it?

source/mir/random/discrete.d
WebDrake, line 75: I would accept it wrapped in a `debug` block, it's having it as part of the regular public API that I'm not happy with.

source/mir/random/discrete.d
9il, line 68: We are going to implement dramatically better FP generators. Description is too large and will be published in article. But all this will be after Tinflex.

source/mir/random/discrete.d
WebDrake, line 91: Hmmm, to me this feels like working around the lack of guarantees from upstream.  Wouldn't it be safer to hold off declaring `opCall` `const` until upstream is fixed?

source/mir/random/discrete.d
9il, line 68: > Not the point -- designing the API this way it will complicating the use of Discrete as a drop-in replacement for existing phobos dice (and also throw on the end user the responsibility of summing proportions correctly).
1. `dice` support is not the goal.
2. This is one/two UFCS calls

source/mir/random/discrete.d
9il, line 92: > Yes I will try to fix Phobos, but that won't be released before 2.073 :/

You have not time to fix Phobos templates, please consider on related issues

source/mir/random/discrete.d
9il, line 52: I just did not know about it

source/mir/random/discrete.d
WebDrake, line 92: I'm not sure that it's appropriate to prioritize putting `const` on `opCall` above having clean and easy-to-follow code.  Things like the `cast` and the two different internal variables for the cumulative distribution points create unnecessary complication in the code, and the `cast` in any case risks breaking the `const` guarantee (certainly if e.g. at some point you generalize the type which can be provided as input to the constructor).

This is a case where the fix needs to be upstream, and where the guarantee wanted here can easily be added after that fix is complete.

source/mir/random/discrete.d
WebDrake, line 58: > OK, then use assumeSorted instead.

This feels like using an inferior implementation for the sake of a public method that isn't strictly necessary for what `Discrete` does.

If the `cdPoints` method is really, really wanted, is it problematic to just return `const(SortedRange)` ... ?  That should serve all the debug use-cases.

source/mir/random/discrete.d
WebDrake, line 68: > We are going to implement dramatically better FP generators. Description is too large and will be published in article.

Cool, looking forward to seeing that work.

WebDrake, line 68: > dice support is not the goal.

But it's nice to give people something they can use as an easy replacement.

> This is one/two UFCS calls

If it's one/two UFCS calls, why not provide it here via an alternative constructor?

(Actually I doubt it's necessarily quite so straightforward when it comes to details of doing sums of integral or floating-point numbers correctly, but that's a longer discussion...)

source/mir/random/discrete.d
WebDrake, line 52: Well, if you want to take a look: https://github.com/WebDrake/hap/blob/e25fc15a17e4e439333740d2a9894fc7d3c94012/source/hap/random/distribution.d#L232

Note there are things that won't be useful here (it's a class, it's a forward range...) but it should be straightforward to take most of the good bits and use them with the functor-based approach here.

source/mir/random/discrete.d
9il, line 75: ok
@wilzbach, lets remove this. anyway we are using assumeSorted now

source/mir/random/discrete.d
wilzbach, line 68: > If it's one/two UFCS calls, why not provide it here via an alternative constructor?
> (Actually I doubt it's necessarily quite so straightforward when it comes to details of doing sums of integral or floating-point numbers correctly, but that's a longer discussion...)

I agree that such an interface would be nice - even in Tinflex we have relative `counts` and need to sum over them.

PR #235, opened 2016-06-26
Address comments from #234

wilzbach:
started to adressed comments from #234.
- renamed unittest variable rndGen -> gen
- renamed r -> _cdSortedRange

@9il @WebDrake @joseph-wakeling-sociomantic

As `back` isn't `const`, just using the sorted range doesn't work.
As an option we might save the last element.

### DIFF
--- a/source/mir/random/discrete.d
+++ b/source/mir/random/discrete.d
@@ -52,7 +52,7 @@ Complexity: O(log n) where n is the number of `cdPoints`.
 struct Discrete(T)
     if (isNumeric!T)
 {
-    private const(T)[] _cdPoints;
+    private const(T)[] cdPoints;
 
     /**
     The cumulative density points `cdPoints` are assumed to be sorted and given
@@ -64,13 +64,7 @@ struct Discrete(T)
     */
     this(const(T)[] cdPoints)
     {
-        _cdPoints = cdPoints;
-    }
-
-    /// cumulative density points
-    const(T)[] cdPoints() const @property
-    {
-        return _cdPoints;
+        this.cdPoints = cdPoints;
     }
 
     /// samples a value from the discrete distribution
@@ -86,8 +80,8 @@ struct Discrete(T)
         import std.random : uniform;
         import std.range : assumeSorted;
 
-        T v = uniform!("[)", T, T)(0, _cdPoints[$-1], gen);
-        return _cdPoints.assumeSorted!"a <= b".lowerBound(v).length;
+        T v = uniform!("[)", T, T)(0, cdPoints[$-1], gen);
+        return cdPoints.assumeSorted!"a <= b".lowerBound(v).length;
     }
 }
 
@@ -113,7 +107,6 @@ unittest
     // 1, 2, 1
     auto cdPoints = [1, 3, 4];
     auto ds = discrete(cdPoints);
-    assert(ds.cdPoints == cdPoints);
 
     auto obs = new uint[cdPoints.length];
     foreach (i; 0..1000)
--- a/source/mir/random/discrete.d
+++ b/source/mir/random/discrete.d
@@ -52,10 +52,7 @@ Complexity: O(log n) where n is the number of `cdPoints`.
 struct Discrete(T)
     if (isNumeric!T)
 {
-    import std.range : SortedRange;
-
     private const(T)[] _cdPoints;
-    private SortedRange!(const(T)[], "a <= b") r;
 
     /**
     The cumulative density points `cdPoints` are assumed to be sorted and given
@@ -68,7 +65,6 @@ struct Discrete(T)
     this(const(T)[] cdPoints)
     {
         _cdPoints = cdPoints;
-        r = typeof(r)(_cdPoints);
     }
 
     /// cumulative density points
@@ -88,15 +84,17 @@ struct Discrete(T)
     size_t opCall(RNG)(ref RNG gen) const
     {
         import std.random : uniform;
+        import std.range : assumeSorted;
+
         T v = uniform!("[)", T, T)(0, _cdPoints[$-1], gen);
-        return (cast(SortedRange!(const(T)[], "a <= b")) r).lowerBound(v).length;
+        return _cdPoints.assumeSorted!"a <= b".lowerBound(v).length;
     }
 }
 
 unittest
 {
     import std.random : Random;
-    auto rndGen = Random(42);
+    auto gen = Random(42);
 
     // 10%, 20%, 20%, 40%, 10%
     auto cdPoints = [0.1, 0.3, 0.5, 0.9, 1];
@@ -104,13 +102,13 @@ unittest
 
     auto obs = new uint[cdPoints.length];
     foreach (i; 0..1000)
-        obs[ds(rndGen)]++;
+        obs[ds(gen)]++;
 }
 
 unittest
 {
     import std.random : Random;
-    auto rndGen = Random(42);
+    auto gen = Random(42);
 
     // 1, 2, 1
     auto cdPoints = [1, 3, 4];
@@ -119,5 +117,5 @@ unittest
 
     auto obs = new uint[cdPoints.length];
     foreach (i; 0..1000)
-        obs[ds(rndGen)]++;
+        obs[ds(gen)]++;
 }


source/mir/random/discrete.d
```diff
@@ -67,8 +67,9 @@ struct Discrete(T)
     */
     this(const(T)[] cdPoints)
     {
+        import std.range : assumeSorted;
         _cdPoints = cdPoints;
-        r = typeof(r)(_cdPoints);
+        _cdSortedRange = _cdPoints.assumeSorted!"a <= b";
```
wilzbach: verify sorted occurs after the SortedRange constructor has been called:
https://github.com/dlang/phobos/blob/master/std/range/package.d#L7919

[`assumeSorted`](https://github.com/dlang/phobos/blob/master/std/range/package.d#L8514) is:

``` d
auto assumeSorted(alias pred = "a < b", R)(R r)
{
    return SortedRange!(Unqual!R, pred)(r);
}
```

so imho I would keep the direct constructor call..

WebDrake: I still feel it's an unnecessary complication of the implementation to have both `_cdPoints` and `_cdSortedRange`.  The latter is all that's actually _needed_, is there anything else that isn't just workaround for the lack of `const` support?

source/mir/random/discrete.d
```diff
@@ -68,14 +68,8 @@ struct Discrete(T)
...
-        _cdPoints = cdPoints;
-        _cdSortedRange = _cdPoints.assumeSorted!"a <= b";
-    }
-
-    /// cumulative density points
-    const(T)[] cdPoints() const @property
-    {
-        return _cdPoints;
+        _cdSum = cdPoints[$ - 1];
+        _cdSortedRange = cdPoints.assumeSorted!"a <= b";
```
WebDrake: Yea, this also works.  It's still a workaround, though ;-)

OTOH it might have some application if e.g. you wanted to generalize the constructor to take an arbitrary `SortedRange` as input (which would probably have to be guaranteed to be a forward range, but not _necessarily_ a bidirectional range).

source/mir/random/discrete.d
```diff
@@ -68,7 +65,6 @@ struct Discrete(T)
     this(const(T)[] cdPoints)
     {
         _cdPoints = cdPoints;
-        r = typeof(r)(_cdPoints);
```
WebDrake, line 71: Maybe something to do in a separate PR, but I think that `cdPoints` can be generalized to a forward range, and it should probably be `save`d to make sure it belongs to the `Discrete` distribution (cf. my point in the earlier issue that `const(T)[]` gives insufficient guarantees that your distribution will not change over time).

The use of `assumeSorted` in `opCall` should at least provide better debug checks that the fundamental property of sortedness isn't being violated, though.

9il:
LGTM

source/mir/random/discrete.d
```diff
@@ -88,29 +84,31 @@ struct Discrete(T)
     size_t opCall(RNG)(ref RNG gen) const
     {
         import std.random : uniform;
+        import std.range : assumeSorted;
+
         T v = uniform!("[)", T, T)(0, _cdPoints[$-1], gen);
```
wilzbach: @WebDrake there are (at least) two ways to do this:

#### inclusive uniform interval & less sortedness

```d
T v = uniform!"[]"(0, cdPoints[$-1], gen);
return cdPoints.assumeSorted!("a < b").lowerBound(v).length;
```
#### semi-inclusive uniform interval & less-or-equal sortedness

```d
T v = uniform!"[)"(0, cdPoints[$-1], gen);
return cdPoints.assumeSorted!("a <= b").lowerBound(v).length;
```

source/mir/random/discrete.d
```diff
@@ -88,36 +78,37 @@ struct Discrete(T)
     size_t opCall(RNG)(ref RNG gen) const
     {
         import std.random : uniform;
-        T v = uniform!("[)", T, T)(0, _cdPoints[$-1], gen);
-        return (cast(SortedRange!(const(T)[], "a <= b")) r).lowerBound(v).length;
+        import std.range : assumeSorted;
+
+        T v = uniform!("[)", T, T)(0, cdPoints[$-1], gen);
```
wilzbach, line 83: @WebDrake there are (at least) two ways to do this:

#### inclusive uniform interval & less sortedness

``` d
T v = uniform!"[]"(0, cdPoints[$-1], gen);
return cdPoints.assumeSorted!("a < b").lowerBound(v).length;
```

#### semi-inclusive uniform interval & less-or-equal sortedness

``` d
T v = uniform!"[)"(0, cdPoints[$-1], gen);
return cdPoints.assumeSorted!("a <= b").lowerBound(v).length;
```

wilzbach:
@WebDrake 

> Maybe something to do in a separate PR, but I think that cdPoints can be generalized to a forward range, and it should probably be saved to make sure it belongs to the Discrete distribution (cf. my point in the earlier issue that const(T)[] gives insufficient guarantees that your distribution will not change over time).

[lowerBound](https://github.com/dlang/phobos/blob/master/std/range/package.d#L8127) (aka binary search) requires at least slicing, the uniform generator needs access to the last element.
Doing a linear search in the case would result in a different complexity with the same syntax which isn't intuitive, so I wouldn't be a huge fan of this.

> The use of assumeSorted in opCall should at least provide better debug checks that the fundamental property of sortedness isn't being violated, though.

check is in `SortedRange`, now the verification of the sortedness is called on every `opCall` instead of being called in the `Discrete` constructor.

source/mir/random/discrete.d
WebDrake, line 83: Am I right in thinking that `a < b` doesn't allow for any discrete element to have zero probability, whereas `a <= b` does ... ?

WebDrake:
> lowerBound (aka binary search) requires at least slicing, the uniform generator needs access to the last element.
> Doing a linear search in the case would result in a different complexity with the same syntax which isn't intuitive, so I wouldn't be a huge fan of this.

Yea, fair points.  What I'm getting at, though, is that you should probably not constrain your user to giving you an array.

> check is in SortedRange, now the verification of the sortedness is called on every opCall instead of being called in the Discrete constructor.

Yes, indeed.  But can it guarantee for you that someone else is not modifying the elements of that array without affecting sortedness?

9il:
> lowerBound (aka binary search) requires at least slicing, the uniform generator needs access to the last element.
> Doing a linear search in the case would result in a different complexity with the same syntax which isn't intuitive, so I wouldn't be a huge fan of this.
> Yea, fair points. What I'm getting at, though, is that you should probably not constrain your user to giving you an array.

Current version is OK. I don't want Mir be another Phobos with template hell

> check is in SortedRange, now the verification of the sortedness is called on every opCall instead of being called in the Discrete constructor.
> Yes, indeed. But can it guarantee for you that someone else is not modifying the elements of that array without affecting sortedness?

No. We have not such goals. Goals are 1. quality 2. speed 3. Small API 4. Clean code

source/mir/random/discrete.d
wilzbach, line 83: I think elements with zero probs are problematic in both cases - see these two examples:

## General example

``` d
import std.range;
auto cdPoints = [1, 3, 4];
foreach (el; 0..5)
    writeln(el, "->", cdPoints.assumeSorted!"a < b".lowerBound(el));
foreach (el; 0..5)
    writeln(el, "->", cdPoints.assumeSorted!"a <= b".lowerBound(el));
```

```
0->[]
1->[]
2->[1]
3->[1]
4->[2] // only included in "[]"

0->[]
1->[1]
2->[1]
3->[1, 3]
4->[1, 3] // only included in "[]"
```

## With zero probs

``` d
import std.range;
auto cdPoints = [0, 1, 3, 4];
foreach (el; 0..5)
    writeln(el, "->", cdPoints.assumeSorted!"a < b".lowerBound(el));
foreach (el; 0..5)
    writeln(el, "->", cdPoints.assumeSorted!"a <= b".lowerBound(el));
```

```
0->[]
1->[0]
2->[0, 1]
3->[0, 1]
4->[0, 1, 3] // only included in "[]"

0->[0] 
1->[0, 1]
2->[0, 1]
3->[0, 1, 3]
4->[0, 1, 3, 4]  // only included in "[]"
```

WebDrake:
> Current version is OK. I don't want Mir be another Phobos with template hell [... ] Goals are 1. quality 2. speed 3. Small API 4. Clean code

OK, fair enough.  It's certainly better to press on with the next challenge rather than chase these questions further now.

9il:
@WebDrake We have a lot of really important issues in Tinflex implementation. They include optimisation and complexity issues. Also a lot of bugs are hidden and may be found during code review. I spent all Saturday to review and wrote comments on Tinflex. As soon as @wilzbach will push the next revision please start implementation review. Implementation is much more important for now and we have really small amount of time to create good Tinflex.

WebDrake:
OK, I'll try to get onto it soon.

source/mir/random/discrete.d
wilzbach, line 83: see #236 for a fix ;-)

PR #236, opened 2016-06-26
mir.random.discrete - fix for zero probilities

wilzbach:
(in reference to #235)

As @WebDrake pointed out `Discrete` doesn't handle zero probabilities well. 
Also as we don't use chi-squared tests, but I think it's at least helpful to insert the assert's for the currently tested distributions.

With zero probs we can't use `lowerBound`, because it will it will not stop on the first element, but go to all equal element. Hence we either count the first category twice or the third one.
Here's an example of what's going on:

``` d
import std.stdio;
import std.range;
auto cdPoints = [1, 3, 3, 4];
foreach (el; 0..5)
{
    writeln(el, "->", cdPoints.assumeSorted!"a < b".lowerBound(el));
}

foreach (el; 0..5)
    writeln(el, "->", cdPoints.assumeSorted!"a <= b".lowerBound(el));
```

```
0->[]
1->[] // would count first category twice
2->[1]
3->[1]
4->[1, 3, 3] // only for "[]"

0->[]
1->[1]
2->[1]
3->[1, 3, 3] // would count third category (the zero one)
4->[1, 3, 3, 4] //only for "[]"
```

### DIFF
--- a/source/mir/random/discrete.d
+++ b/source/mir/random/discrete.d
@@ -37,7 +37,7 @@ unittest
 
     // sample from the discrete distribution
     auto obs = new uint[cdPoints.length];
-    foreach (i; 0..1000)
+    foreach (i; 0..10_000)
         obs[ds()]++;
 }
 
@@ -81,10 +81,11 @@ struct Discrete(T)
         import std.range : assumeSorted;
 
         T v = uniform!("[)", T, T)(0, cdPoints[$-1], gen);
-        return cdPoints.assumeSorted!"a <= b".lowerBound(v).length;
+        return cdPoints.length - cdPoints.assumeSorted!"a < b".upperBound(v).length;
     }
 }
 
+// test with cumulative probs
 unittest
 {
     import std.random : Random;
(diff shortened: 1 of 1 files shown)


source/mir/random/discrete.d
```diff
@@ -37,7 +37,7 @@ unittest
 
     // sample from the discrete distribution
     auto obs = new uint[cdPoints.length];
-    foreach (i; 0..1000)
+    foreach (i; 0..10_000)
```
wilzbach, line 40: the split looks too random with `1000`

source/mir/random/discrete.d
```diff
@@ -109,6 +113,25 @@ unittest
...
+        obs[ds(gen)]++;
+
+    assert(obs == [2536, 4963, 2501]);
+}
+
+// test with zero probabilities
+unittest
+{
+    import std.random : Random;
+    auto gen = Random(42);
```
9il: OK, but don not use `Random` alias then. Use exact name instead.

WebDrake: Agree, it's better to specify an exact generator just in case someone later adds some tests that depend on the pseudo-random sequence.

wilzbach:
> OK, but don not use Random alias then. Use exact name instead.

Ok - done ;-)

Issue #237, opened 2016-06-26
Discrete distribution sampling in O(1) using Alias or Marsaglia method


wilzbach:
Both the [Alias method](https://en.wikipedia.org/wiki/Alias_method) and [Marsaglia method](https://www.jstatsoft.org/article/view/v011i03) allow to sample from a discrete distribution in O(1).

(related to #233)

9il:
yep, Discrete should be renamed to `NaiveDiscrete` thought. And Alias method should be added.

PR #240, opened 2016-06-27
Tinflex

wilzbach:
(reboot of #222 with an upstream branch)
# TODO
- [x] Move  `LinearFun` to `mir.utilitiy.linearfun`, make it public. Rename  `_y` -> `x0`, `_a`->`a` (remove `a` property). Remove `opEquals` (should work out of the box). Clean and add docs.
- [x] Move `GenerationInterval` to `mir.tinflex.package`, make it public and rename to `TinflexInterval`
- [x] Rename `calcPoints` -> `tinflexIntervals` and move it to `mir.tinflex.package`.
- [x] `tinflex` should have following API:
  1. `(in Pdf pdf, in F0 f0, in F1 f1, in F2 f2, S[] cs, S[] points, S rho = 1.1)`
  2. `(in F0 f0, in F1 f1, in F2 f2, S[] cs, S[] points, S rho = 1.1)`
  3. `(in Pdf pdf, in TinflexInterval!S[] intervals)`
- [x] `tinflexImpl` should have API: `(in Pdf pdf, in GenerationInterval!S[] gvs, in Discrete!S ds, ref RNG rng)`

after that users (and you) would be able to analyse tinflex intervals without dirty hacks like `plot.d`. So, create a single file in examples with all test like in R code.
- [x] Use matplotlib instead of ggplotd
- [x] Remove ggplotd
- [x] inline `transform` and remove it.
- [x] Interval lines should look exactly like in R. `inverse` should be used, thus lines would be curved.
- [x] Use http://pyd.readthedocs.io/en/latest/functions.html to pass interval function to python instead of computing dots in D
- [x] Remove all unused stuff from `mir.tinflex`
- [x] rename `inverse` to `flexInverse`, move to package, make public
- [x] Rename from `tinflex` to `flex` (all cases)
- [x] replace stuff like `z / 3` with `z * one_div_3`
- [x] fix `dub.json`: package must not define configurations for examples. Use new builtin dub syntax, but with `dub.json` instead of `dub.sdl` (see http://code.dlang.org/getting_started, Single-file packages)
- [x] Do detailed paper work for cases +- 1 in Table 3. Need z^3 approximation instead of z^2. 
- [x] Do detailed paper work for Table 3 analog but with right bound instead of left. This will fix sampling for -infinity (btw, how -infinity handled in R code?)
- [x] Fix bugs with in `tinflexImpl`
- [x] Rename `Discrete` -> `NaiveDiscrete`
- [x] Add Alias Method as `Discrete`
- [x] Comment article https://www.academia.edu/25911888/An_automatic_code_generator_for_nonuniform_random_variate_generation?auto=download
  - they provide a tool to generate the hat/squeeze values for various languages, combined with a generated sampling method based on the hat/squeeze values
  - hence no additional runtime cost is required
  - tool is no longer available online
- [x] Use binary heap (and sort in the end) instead of Dlist
- [x] clean code, add docs
- [x] look at FP failures on x86
- [x] remove commented tests


wilzbach:
@WebDrake @joseph-wakeling-sociomantic just a short update, after a lot of help from @9il (you are awesome!), and many fixed bugs (my favorite failure one is the format in which I save - see [the wiki](https://github.com/libmir/mir/wiki/Numerical-failures) for more details), the Tinflex algorithm is approaching stability (there are still some open issues - see below). In any case there is a lot to say, but in brevity too small highlights:

1) `area.d` and `transformation.d` have now an extensive test data. It was generated by running (parts of) the Tinflex algorithm in R.
2) The standard normal distribution construction works (however the Tinflex algorithm requires every  unbounded interval to be concave and strictly monotone, so `double.infinity` does work as point)

I think the best example of the progress can be shown by this output from the Tinflex algorithm.
It constructs a hat and squeeze function for the standard normal distribution with a user specified efficiency `rho`.

`rho` = 1.1 (6 intervals)

image

`rho` = 1.02 (10 intervals)

image

`rho` = 1.005 (18 intervals)

image

`rho` = 1.0005 (70 intervals)

image

There is no real limit how far the efficiency can be tuned, e.g. `rho` = 1.00005 requires 218 intervals.
As the "improved ziggurat" method has an efficiency of `1.02`, it will be very interesting to benchmark those ;-)

Moreover I ported the Tinflex tests from R to check whether everything is working. Here's the current status:

## Current status of the example tests

1) `test4` produces some `NaN` intervals -> fails with range exception during sampling
2) most hat, squeeze plots look good, except for
- hat/squeeze & histogram in dist1c (`[-double.infinity, -1, 0, 1, double.infinity]`) - dist1b is also unbounded, but seems to work
-  right part of hat/squeeze in dist3 (most generated point are `NaN`)

3) Histograms seem to be worse
- dist1_b-0.9 (other c here seem to work
- dist1e
- dist2_c_0.5
- dist3 (completly, probably due to wrong hat/squeeze)
- normal distribution





## What else is missing
- [ ] prettify and customize plotting for users (currently it heavily relies on `ggplotd`)
- [ ] Alias or Magsalia method for sampling in O(1) (#237)
- [ ] Code cleanup (the last days involved a lot of refactoring - this is still "noticeable")
- [ ] More extensive user documentation (current intro text isn't very helpful to the user, it can be previewed [here](http://preview.mir.rocks/mir_random_tinflex.html))

Edit: I forgot to mention that
- test: `dub test`
- plot all `dub -c plot-generic`
- plot a single `dub -c plot-generic -- normal` or `dub -c plot-generic -- 1`

It currently produces a lot of logs, you can remove the `Tinflex_logging` version in the `dub.json` to disable this.

source/mir/random/tinflex/internal/area.d
```diff
@@ -0,0 +1,650 @@
...
+    import mir.random.tinflex.internal.linearfun : emptyFun, secant, tangent;
+    import mir.random.tinflex.internal.types : determineType, FunType;
+
+    enum sec = "secant(iv.lx, iv.rx, iv.ltx, iv.rtx)";
+    enum t_l = "tangent(iv.lx, iv.ltx, iv.lt1x)";
+    enum t_r = "tangent(iv.rx, iv.rtx, iv.rt1x)";
+
+    // could potentially be saved for subsequent calls
+    FunType type = determineType(iv);
+    with(FunType) with(iv)
```
joseph-wakeling-sociomantic: Minor: I would drop the `with(iv)` here; I think the repetitiveness of writing `iv.squeeze`, `iv.hat` everywhere is worth it for the clarity of what's being written to.

Personally I'd drop the `with(FunType)` too, but that's less of an issue.

source/mir/random/tinflex/internal/area.d
```diff
@@ -0,0 +1,650 @@
...
+
+    (F_T(sh(r))- F_T(sh(l))) / sh.slope
+
+Params:
+    sh = linear function
+    l  = start of interval
+    r  = end of interval
+    ly = start of interval (y-value)
+    ry = end of interval (y-value)
+    c  =  interval type (see paper)
```
joseph-wakeling-sociomantic: Params don't match the actual input parameter.  Other function description should probably also be rewritten accordingly.

source/mir/random/tinflex/internal/area.d
```diff
@@ -0,0 +1,650 @@
+module mir.random.tinflex.internal.area;
+
+import mir.random.tinflex.internal.types : Interval;
+import std.traits : ReturnType;
+
+/**
+Determines the hat and squeeze function of an interval.
+Based on Theorem 1
```
joseph-wakeling-sociomantic: I would always give a minimal citation of the paper, i.e. here `Theorem 1 of Botts et al. (2013)`.  The full citation should be given in module documentation.

joseph-wakeling-sociomantic: Also, missing `Template_Params:` and `Params:` sections.

source/mir/random/tinflex/internal/area.d
```diff
@@ -0,0 +1,650 @@
...
+                squeeze = squeeze.init;
+                hat = mixin(t_r);
+                break;
+            }
+            if (iv.rx == +S.infinity)
+            {
+                squeeze = squeeze.init;
+                hat = mixin(t_l);
+                break;
+            }
```
joseph-wakeling-sociomantic: Where do these 2 special cases come from?  Probably worth adding a comment, since they are not obvious just from Theorem 1.

source/mir/random/tinflex/internal/area.d
```diff
@@ -0,0 +1,650 @@
...
+            squeeze = mixin(sec);
+            hat = iv.ltx > iv.rtx ? mixin(t_l) : mixin(t_r);
+            break;
+        case T4b:
+            squeeze = iv.ltx < iv.rtx ? mixin(t_l) : mixin(t_r);
+            hat = mixin(sec);
+            break;
+        default:
+            squeeze = emptyFun!S;
+            hat = emptyFun!S;
```
joseph-wakeling-sociomantic: Is this `default:` a legitimate case in any circumstances?  If not, it might be better just to `assert(false)`.

source/mir/random/tinflex/internal/area.d
```diff
@@ -0,0 +1,650 @@
...
+        auto hs2 = dhs(1, 3);
+        assert(hs2.hat.slope == 4);
+        assert(hs2.hat.intercept == -3);
+        assert(hs2.squeeze.slope == 2);
+        assert(hs2.squeeze.intercept == -1);
+    }
+}
+
+alias hatArea(S) = area!(true, S);
+alias squeezeArea(S) = area!(false, S);
```
joseph-wakeling-sociomantic: These aliases should be individually documented, at least if they are intended to be public.

source/mir/random/tinflex/internal/area.d
```diff
@@ -0,0 +1,650 @@
...
+
+/**
+Computes the area below a function sh in-between l and r.
```
joseph-wakeling-sociomantic: Isn't the area below a function just the definite integral?  Whatever you call the function, might be worth making the connection.

source/mir/random/tinflex/internal/area.d
```diff
@@ -0,0 +1,650 @@
...
+    sh = linear function
+    l  = start of interval
+    r  = end of interval
+    ly = start of interval (y-value)
+    ry = end of interval (y-value)
+    c  =  interval type (see paper)
+
+Returns: Computed area below sh.
+*/
+void area(bool isHat, S)(ref Interval!S iv)
```
joseph-wakeling-sociomantic: Missing `Template_Params:` documentation.

joseph-wakeling-sociomantic: The `bool` template parameter could be better achieved with an `enum bool` that has values `squeeze, hat`.  It would make your calls to `area` better self-documenting.

source/mir/random/tinflex/internal/area.d
```diff
@@ -0,0 +1,650 @@
...
+        if (area == S.max)
+        {
+            area = 0;
+        }
+    }
+
+    static if (isHat)
+        iv.hatArea = area;
+    else
+        iv.squeezeArea = area;
```
joseph-wakeling-sociomantic: These `static if` statements related to `isHat` can probably be combined into one `static if/else` block.

9il: this is just more clear how it is

source/mir/random/tinflex/internal/area.d
9il: they are not

source/mir/random/tinflex/internal/area.d
9il: `assert(false)` will halt the program

source/mir/random/tinflex/internal/area.d
9il: all this code is in internal folder and not part of public api

source/mir/random/tinflex/internal/area.d
joseph-wakeling-sociomantic: OK, in that case they should be marked `private` (or `package` if they're needed by other tinflex-related modules).

source/mir/random/tinflex/internal/area.d
joseph-wakeling-sociomantic: It's currently a public function, though.

source/mir/random/tinflex/internal/area.d
joseph-wakeling-sociomantic: Assuming that `default:` is meant to be an unreachable case, isn't that desirable?

source/mir/random/tinflex/internal/area.d
```diff
@@ -0,0 +1,650 @@
...
+        immutable z = leftOrRight * sh.slope * (iv.rx - iv.lx);
+        // check whether approximation is possible, page 5
+        if (abs(z) < S(1e-6))
+        {
+            area = exp(sh.a) * (iv.rx - iv.lx) * (1 + z / 2 + (z * z) / 6);
+        }
+        else
+        {
+            // F_T = e^x
+            area = (exp(sh(iv.rx)) - exp(sh(iv.lx))) / sh.slope;
```
joseph-wakeling-sociomantic: Is this coming from Equation (3) with `x0 == 0` .... ?

source/mir/random/tinflex/internal/area.d
wilzbach: > The bool template parameter could be better achieved with an enum bool that has values squeeze, hat. It would make your calls to area better self-documenting.

this is only the internal API and both `hatArea` and `squeezeArea` are aliased (see ten lines above):

```
alias hatArea(S) = area!(true, S);
alias squeezeArea(S) = area!(false, S);
```

The `area` function is only acesed via those two aliases, so I think that's pretty clear?

wilzbach:
Just a quick heads-up on the hat/squeeze for dist3.

1) I found a very tricky issue between the tests of it in `area.d` and `calc.d`
("magic difference between 0.0 (`0x0p+0`) and -0.0 (in this case `-0x8p-70`))

I put a reduced version of this behavior in a [gist](https://gist.github.com/wilzbach/ddc216a90c7a86964c637d676f17a32c) if you want to play with it.

It turns out this doesn't matter for the Tinflex algorithm as it just splits the interval until this weird numeric issue with a negative zero doesn't happen.

The underlying problem is that the derivative of distribution 3 is `NaN` for `-1` and `1` and thus the left-most and right-most interval should be `0`.

```
auto f0 = (S x) => log(1 - x^^4);
auto f1 = (S x) => -4 * x^^3 / (1 - x^^4);
```

image

This means our transformed transformed density (orange) looks like this.
In this case (type IVb), for the hat function as from a secant between -1.0 and -0.9 (red) or 0.9 and 1.0 (green) is constructed:

image

I am still not entirely sure why we get the "negative zero" only in some specific cases (see above), but at the beginning of `area.d` we check against such weird cases and set the area to infinity, which will trigger a split, where the difference between left and right is less and thus the risk of the numerical errors is reduced.

``` d
if (iv.c * sh(iv.lx) < 0 || iv.c * sh(iv.rx) < 0)
{
    area = S.max;
    goto L;
```

The `c` here is necessary as our transformation function t is `exp(c * f0)`.
So in summary I hope that this numeric problem doesn't have any measurable effects on the distribution, but as long as we can construct the `hat` and `squeeze` functions, it shouldn't matter whether 6 or 8 Intervals are needed to represent the distribution.

2) On a brighter note I found out why the hat/squeeze plots for dist3 looked so weird. It was just a stupid less (`while (xs[k] < r)`) instead of an less-equal and a big step-size of `0.1`. Fixing the `less`, increasing the efficiency `rho` and using a `0.01 step-size already produces a better plot:

image

Note bene: there's still a small weird artificat at the right, but that looks like a plotting error.

source/mir/random/tinflex/internal/area.d
```diff
@@ -0,0 +1,700 @@
...
+import std.traits : ReturnType;
+
+/**
+Determines the hat and squeeze function of an interval.
+Based on Theorem 1
+*/
+void determineSqueezeAndHat(S)(ref Interval!S iv)
+in
+{
+    assert(iv.lx < iv.rx, "invalid interval");
```
joseph-wakeling-sociomantic: Isn't this something that could be guaranteed by `Interval` ... ?

source/mir/random/tinflex/internal/area.d
joseph-wakeling-sociomantic: > this is only the internal API and both hatArea and squeezeArea are aliased

OK, fair enough.

source/mir/random/tinflex/internal/area.d
9il: the problem that no one from us has full picture of what FP behaviour can be in the functions.

wilzbach:
> the problem that no one from us has full picture of what FP behaviour can be in the functions.

A good example is this [gist](https://gist.github.com/wilzbach/d86ddb98aaaaef5531898b457f8e2a7f) here. It's derived from the example above, but just shows the weirdness of using different FP types -  `float` < 0, `double` > 0, `real` < 0.

source/mir/random/tinflex/internal/area.d
PetarKirov: @joseph-wakeling-sociomantic BTW, I think `Template_Params` works only in D1. (AFAIR it was added last year.) In D2 `Params` is used for both template and runtime parameters.

wilzbach:
Two quick questions:

1) Public API

> tinflex should have following API:

``` d
(in Pdf pdf, in F0 f0, in F1 f1, in F2 f2, S[] cs, S[] points, S rho = 1.1)
(in F0 f0, in F1 f1, in F2 f2, S[] cs, S[] points, S rho = 1.1)
(in F0 f0, in F1 f1, in F2 f2, in TinflexInterval!S[] intervals)
(in Pdf pdf, in TinflexInterval!S[] intervals)
```

`Pdf` is the same as `F0`, I guess we should we also provide a mean to set `maxIntervals`?

``` d
 // one global c is the default case
(in Pdf pdf, in F1 f1, in F2 f2, S cs, S[] points, S rho = 1.1, size_t maxIntervals = 1_000)
(in Pdf pdf, in F1 f1, in F2 f2, S[] cs, S[] points, S rho = 1.1, size_t maxIntervals = 1_000)
(in Pdf pdf, in TinflexInterval!S[] intervals)
```

2) Plotting requires the `inverse` function and if the transformed plot is shown, the `transformation` function. Should we expose this implementation details to the user or should we better make the abstraction `plot` that can be called by the user?

source/mir/random/tinflex/internal/area.d
wilzbach: > Minor: I would drop the `with(iv)` here; 

Hmm to be honest I am a huge fan of this style as it removes the repetitiveness from the `switch`. @9il what do you think?

9il:
> Pdf is the same as F0, I guess we should we also provide a mean to set maxIntervals?

F0(x) = is log(Pdf(x)), if i am not wrong.

Yes, first 2 prototypes should have `approxMaxIntervals`.

> 2) Plotting requires the inverse function and if the transformed plot is shown, the transformation function. Should we expose this implementation details to the user or should we better make the abstraction plot that can be called by the user?

Use `tinflexIntervalTransformation` public function name. Do not add API for plotting, but interval may have `opCall` function that uses the transformation.

source/mir/random/tinflex/internal/area.d
```diff
@@ -0,0 +1,669 @@
...
+            {
+                squeeze = squeeze.init;
+                hat = mixin(t_l);
+                break;
+            }
+            squeeze = mixin(sec);
+            hat = iv.ltx > iv.rtx ? mixin(t_l) : mixin(t_r);
+            break;
+        case T4b:
+            squeeze = iv.ltx < iv.rtx ? mixin(t_l) : mixin(t_r);
```
wilzbach: From the Tinflex paper (page 3 - left side after the first listing)

> t_m is the tangent of f in bl or br wherever f is larger

wilzbach: current code:

image

after changing it according to the paper:

image

wilzbach: with rho = 1.05

left is t_m according to the paper



with rho = 1.01

source/mir/random/tinflex/internal/area.d
wilzbach: > OK, in that case they should be marked private (or package if they're needed by other tinflex-related modules).

@9il's policy is to make everything public for hackers, but state with `internal` that the API isn't supported

@9il can I move them back down so that the aliases are closer to the `area` function?

source/mir/random/tinflex/internal/area.d
9il: matplotlib charts with proper transformation should be created before we start to discuss what variant is correct. Plotting should look the same way like in R.

source/mir/random/tinflex/package.d
```diff
@@ -166,7 +167,22 @@ struct Tinflex(Pdf, S)
 
     S pdf(S x) @property const
     {
-        return _pdf(x);
+        import mir.internal.math : exp;
+        bool isLog = true;
```
9il: I wrote you to remove `pdf` method completely.

examples/tinflex_plot.d
```diff
@@ -275,14 +275,17 @@ void test6(string folderName)
...
     //S[] points = [-S.infinity, -1.5, 0, 1.5, S.infinity];
     S[] points = [-3, -1.5, 0, 1.5, 3];
     S sqrt2PI = sqrt(2 * PI);
-    auto f0 = (S x) => 1 / (exp(x * x / 2) * sqrt2PI);
-    auto f1 = (S x) => -(x/(exp(x * x/2) * sqrt2PI));
-    auto f2 = (S x) => (-1 + x * x) / (exp(x * x/2) * sqrt2PI);
+    //auto f0 = (S x) => 1 / (exp(x * x / 2) * sqrt2PI);
+    //auto f1 = (S x) => -(x/(exp(x * x/2) * sqrt2PI));
+    //auto f2 = (S x) => (-1 + x * x) / (exp(x * x/2) * sqrt2PI);
+    auto f0 = (S x) => log(1 / (exp(x * x / 2) * sqrt2PI));
```
9il: reduce formala

source/mir/random/tinflex/package.d
wilzbach: `pdf` is needed for plotting and it also allows to distinguish whether a logarithmic pdf or normal pdf was provided

source/mir/random/tinflex/internal/calc.d
```diff
@@ -0,0 +1,154 @@
...
+ \int^l^r      h(x) - s(x)  dx
+
+It can't be applied for unbounded intervals if c <= -1/2
+
+References:
+    Hormann, W., J. Leydold, and G. Derflinger.
+    "Automatic Nonuniform Random Number Generation." (2004): Formula 4.22
+*/
+auto expmean(S)(in ref Interval!S iv)
+out (result)
```
wilzbach: while looking for the correct reference of "arc-mean", I found the exp-mean (exp_expected new construction point) method.

image

I went ahead and implemented it in a very naive way to test its implications (please no performance critic yet).

However for the NormalDistribution with arcmean 34 intervals are required, with exp-mean 46 :/

`h(x)` and  `s(x)` are defined as linear functions with three parameters `f(x) = a * b * (x - y)`.
I checked my result with WolframAlpha:

upper: http://www.wolframalpha.com/input/?i=Integrate%5B(a_1%2Bb_1(x-y_1)+-+a_2%2Bb_2(x-y_2)),+x%5D

lower: http://www.wolframalpha.com/input/?i=Integrate%5Bx*+(a_1%2Bb_1(x-y_1)+-+a_2%2Bb_2(x-y_2)),+x%5D

So far I couldn't find more references of this method except this tiny excerpt from above, so I hope I interpreted it right.

source/mir/random/tinflex/internal/area.d
wilzbach: > Isn't this something that could be guaranteed by Interval ... ?

It wasn't, but it makes a lot of sense to check this in the invariant. Thanks for the suggestion!

wilzbach:
@9il I added experimental, alpha support for plotting with both Matplotlib and ggplotd. Both can plot a histogram, the intervals as individual lines and the intervals as one consecutive line. Both approaches have advantages, e.g.
- for ggplotd we have the native interface and thus don't need any manual conversion. Also it's a good showcase for D to avoid the dependency to other languages for plotting
- for MatplotLib - well designed, _fast_ & feature-rich library. For example the histogram plotting is quite slow in ggplotd.

```
dub -c tinflex-plot -- norm // plots with Matplotlib
dub -c tinflex-plot-gg -- norm // plots with ggplotd
```

## Intervals as consecutive or as separate lines
- separate: easier to compare distribution with the R output
- consecutive: it's more intuitive + easier to see the efficiency of the intervals as the exact hat and squeeze function are plotted

## Overview

(ggplotd on the left, Matplotlib on the right)

Intervals as lines:

image

Consecutive (aka Area):

image

Histogram: 

(currently invalid)

image

source/mir/random/tinflex/internal/area.d
9il: > for ggplotd we have the native interface and thus don't need any manual conversion. Also it's a good showcase for D to avoid the dependency to other languages for plotting
> for MatplotLib - well designed, fast & feature-rich library. For example the histogram plotting is quite slow in ggplotd.

remove ggplot. Use http://pyd.readthedocs.io/en/latest/functions.html to pass interval function to python instead of computing dots in D

9il:
> for ggplotd we have the native interface and thus don't need any manual conversion. Also it's a good showcase for D to avoid the dependency to other languages for plotting
> for MatplotLib - well designed, fast & feature-rich library. For example the histogram plotting is quite slow in ggplotd.

remove ggplot. Use http://pyd.readthedocs.io/en/latest/functions.html to pass interval function to python instead of computing dots in D

> separate: easier to compare distribution with the R output
> consecutive: it's more intuitive + easier to see the efficiency of the intervals as the exact hat and squeeze function are plotted

Allow only separate.

9il:
And fix your changes in arcmean

9il:
> h(x) and s(x) are defined as linear functions with three parameters f(x) = a \* b \* (x - y).
> I checked my result with WolframAlpha:

they should be defined like `inverse(c, hat(x))`

source/mir/random/flex/internal/calc.d
```diff
@@ -0,0 +1,149 @@
...
+    S y = r;
+    static if (!sorted)
+    {
+        if (y < x)
+        {
+            S t = x;
+            x = y;
+            y = t;
+        }
+    }
```
wilzbach: this doesn't make sense. An `Interval` guarantees `lx < rx`.
@9il can I change the api to a `Interval` or move it somewhere else if it's thought to be more generic?

source/mir/random/tinflex/internal/area.d
wilzbach: btw I found this quote that explains this a bit in the paper:

> Then we may use any tangent with point of contact in [bl,br]. The particular
> choice of tangents deserve some explanation. When ˜f is
> concave and strictly monotone on some interval we always
> use its **maximum** as construction point for the tangent as this
> ensures a valid hat function for f , see Sect. 2.3 below.

source/mir/random/flex/internal/calc.d
9il: Yes

source/mir/random/flex/package.d
```diff
@@ -0,0 +1,925 @@
...
+                S mx2 = f2(mid);
+
+                // apply transformation to right side (for c=0 no transformations are applied)
+                S mt0 = (c == 0) ? mx0 : copysign(S(1), c) * exp(c * mx0);
+                S mt1 = (c == 0) ? mx1 : c * mt0 * mx1;
+                S mt2 = (c == 0) ? mx2 : c * mt0 * (c * pow(mx1, 2) + mx2);
+
+                Interval!S midIP = Interval!S(mid, it.front.rx, it.front.c,
+                                              mt0, mt1, mt2,
+                                              it.front.rtx, it.front.rt1x, it.front.rt2x);
```
wilzbach: @9il I inlined as requested, but won't this make it super hard to experiment with other , custom transformations?

source/mir/random/flex/package.d
```diff
@@ -311,7 +321,7 @@ private S flexImpl(Pdf, S, RNG)
                 }
                 else
                 {
-                    X = hat.y + (log(1 / eXInv + hat.slope * u) - hat.a) / hat.slope;
+                    X = hat.inverse(log(hat.slope * u + exp(hatLx)));
```
9il, line 351: Why? How C code looks like?

source/mir/random/flex/package.d
```diff
@@ -344,8 +354,7 @@ private S flexImpl(Pdf, S, RNG)
                         goto finish;
                     }
                 }
-                // common approximation
-                X = hat.y + (inverseAntiderivative(antiderivative(hatLx, c) + hat.slope * u, c) - hat.a) / hat.slope;
+                X = hat.inverse(inverseAntiderivative(u * hat.slope + antiderivative(hatLx, c), c));
```
9il, line 384: ditto

source/mir/random/flex/package.d
wilzbach, line 351: Proof: http://files.wilzbach.me/mir/reports/tinflex.pdf, page 7

image

....

image

C code: https://github.com/cran/Tinflex/blob/master/src/Tinflex.c#L225

source/mir/random/flex/package.d
wilzbach, line 384: Comes from my [draft report](http://files.wilzbach.me/mir/reports/tinflex.pdf) on page 7.
As mentioned `inverse` is the generalization - please note that the formula for `X` doesn't change here.

source/mir/random/flex/internal/area.d
```diff
@@ -0,0 +1,668 @@
...
+    enum sec = `(iv.ltx >= iv.rtx) ?
+                linearFun!S((iv.rtx - iv.ltx) / (iv.rx - iv.lx), iv.lx, iv.ltx) :
+                linearFun!S((iv.rtx - iv.ltx) / (iv.rx - iv.lx), iv.rx, iv.rtx)`;
+
+    enum t_l = "linearFun!S(iv.lt1x, iv.lx, iv.ltx)";
+    enum t_r = "linearFun!S(iv.rt1x, iv.rx, iv.rtx)";
+
+    // could potentially be saved for subsequent calls
+    FunType type = determineType(iv);
+    with(FunType) with(iv)
```
joseph-wakeling-sociomantic, line 87: I repeat my earlier suggestion that the `with` statements here should be removed.  What they save in typing, they remove in clarity of what's being altered (i.e. there is nothing in the code as written to suggest that `squeeze` and `hat` belong to `iv`).

source/mir/random/flex/internal/area.d
```diff
@@ -0,0 +1,668 @@
+module mir.random.flex.internal.area;
+
+import mir.random.flex.internal.types : Interval;
+import std.traits : ReturnType;
+
+/**
+Determines the hat and squeeze function of an interval.
+Based on Theorem 1 of Botts et al. (2013).
```
joseph-wakeling-sociomantic, line 65: This documentation should make clear that the hat and squeeze functions are set in-place in the `Interval` instance provided to the function.

source/mir/random/flex/internal/area.d
```diff
@@ -0,0 +1,668 @@
...
+                squeeze = squeeze.init;
+                hat = mixin(t_r);
+                break;
+            }
+            if (iv.rx == +S.infinity)
+            {
+                squeeze = squeeze.init;
+                hat = mixin(t_l);
+                break;
+            }
```
joseph-wakeling-sociomantic, line 128: I still think that these `if` special cases should have comments clarifying exactly what part of the Tinflex paper they are described in.

Also, picky, but can we be certain that both `if` conditions won't apply?  It seems to be an implicit assumption of the code here.

source/mir/random/flex/internal/area.d
```diff
@@ -0,0 +1,668 @@
...
+/**
+Computes the area below either the hat or squeeze function
+in-between a interval `iv`.
+Based on table 1 and general equation (3) from Botts et al. (2013).
+
+    (F_T(sh(r))- F_T(sh(l))) / sh.slope
+
+Params:
+    isHat = whether to calculate the hat or squeeze area
+    iv = Interval to use
```
joseph-wakeling-sociomantic, line 212: Missing `Params:` documentation for `S`.

source/mir/random/flex/internal/area.d
```diff
@@ -0,0 +1,668 @@
...
+    // sigma in the paper (1: left side, -1, right side
+    immutable S leftOrRight = (iv.rx - sh.y) > (sh.y - iv.lx) ? 1 : -1;
+    immutable shL = sh(iv.lx);
+    immutable shR = sh(iv.rx);
+    immutable ivLength = iv.rx - iv.lx;
+    auto z = leftOrRight * sh.slope * ivLength;
+
+    // sh.y is the boundary point where f obtains its maximum
+
+    // specializations for T_c family (page 6)
```
joseph-wakeling-sociomantic: This refers to `Table 2`, no?  I would reference it exactly.

source/mir/random/flex/internal/area.d
```diff
@@ -0,0 +1,668 @@
...
+    if (iv.c == 0)
+    {
+        if (fabs(z) < constants!S.smallExp)
+        {
+            area = exp(sh.a) * (iv.rx - iv.lx) * (1 + z * S(0.5) + (z * z) * one_div_6
+                                                    + (z * z * z) * one_div_24);
```
joseph-wakeling-sociomantic: Where does the `z ^^ 3 / 24` factor come from?  It would be good to have a reference or a comment explaining, since it's different from what's on p. 6 (Table 2) of the paper.

source/mir/random/flex/internal/area.d
```diff
@@ -169,7 +172,8 @@ body
     {
         if (fabs(z) < constants!S.smallExp)
         {
-            area = exp(sh.a) * (iv.rx - iv.lx) * (1 + z / 2 + (z * z) / 6 + (z * z * z) / 24);
+            area = exp(sh.a) * (iv.rx - iv.lx) * (1 + z * S(0.5) + (z * z) * one_div_6
```
joseph-wakeling-sociomantic: Can you please explain for me the `exp(sh.a)` part of this formula?

IIUC from Table 2, this is the `f(x0)` component of the area approximation.  I don't follow how it is guaranteed to be `exp(sh.a)` (== exp(alpha) in the paper's notation, right?).

joseph-wakeling-sociomantic: Also, there's some duplication here: you can probably use `ivLength` instead of an explicit `iv.rx - iv.lx`, no ... ?

source/mir/random/flex/internal/area.d
```diff
@@ -186,13 +190,13 @@ body
         }
         else if (iv.c == 1)
         {
-            area = S(0.5) * sh.a * intLength * (z + 2);
+            area = S(0.5) * sh.a * ivLength * (z + 2);
```
joseph-wakeling-sociomantic: Can you again explain the `sh.a` contribution here ... ?  Is it supposed to come from `f(x0)` ... ?

source/mir/random/flex/internal/area.d
wilzbach: > Where does the z ^^ 3 / 24 factor come from?

Further expansion of Taylor series. 

> It would be good to have a reference or a comment explaining,

[Reference is already in work](http://files.wilzbach.me/mir/reports/tinflex.pdf)

source/mir/random/flex/internal/area.d
wilzbach: > Can you please explain for me the exp(sh.a) part of this formula?

[See 2.2](http://files.wilzbach.me/mir/reports/tinflex.pdf). Please note that I get different results as in the paper for the approximation with the "Z-trick". Hence the current `area` code is mostly based on their approximations, but I hope that we can use the results from the small report instead (if we all agree they are correct).

>  exp(sh.a) (== exp(alpha) in the paper's notation, right?).

Yes.

> Also, there's some duplication here: you can probably use ivLength instead of an explicit iv.rx - iv.lx, no ... ?

Yes, thanks.

source/mir/random/flex/internal/area.d
joseph-wakeling-sociomantic: Yea, I think it would be good to have a short comment noting that this expands the Taylor series of the approximation one term more than given in Table 2.  You can also cross-ref any supplementary documentation such as that you are preparing.

source/mir/random/flex/internal/area.d
wilzbach: > Can you again explain the sh.a contribution here ... ? Is it supposed to come from f(x0) ... ?

See page 1 of the report (or point 2.6 in the paper)
1) `x0` can either be `b_l` or `b_r`
2) For both secand and tangent, `a` is defined as `f(x_0)`

joseph-wakeling-sociomantic: On p.1 of the `Explanation` document, you have `beta = g'(x)`.  Do you mean `beta = g'(x0)` ... ?

joseph-wakeling-sociomantic: In the same document, when you write `g(x)` do you mean what in the Tinflex paper is given as `Tc(x)` ... ?

source/mir/random/flex/internal/area.d
wilzbach, line 212: > Missing Params: documentation for S.

Afaik template parameter's aren't usually documented in Phobos if they are obvious from other parameters. Anyway this is still an internal API and the documentation is just for maintainers plus @9il is a bit against adding noise.
I initially used the constraint`isFloatingPoint!S` all over, but he successfully convinced me that to everyone looking in the internals this will be obvious.

source/mir/random/flex/internal/area.d
wilzbach, line 128: > I still think that these if special cases should have comments clarifying exactly what part of the Tinflex paper they are described in.

Sorry that comment got lost. It is based on this information from page (6) (right side, first paragraph):

> In Condition 4 we have demanded that f  ̃ must be concave and strictly monotone in each unbounded interval (−∞, b1] or [bn−1 , ∞) of the given starting partition. 
> ...
> If [bn−1, ∞) contains an inflection point, then it is not possible to construct a hat function according to our rules. Thus Ah,n−1 is set to ∞. 

thanks for reminding. In `types.d` we already have the special case for infinities, which returns either T4a or T4b, hence the check here. However I just realized that we don't have a check for T4b -> cd1d825

source/mir/random/flex/internal/area.d
wilzbach: >  Do you mean beta = g'(x0) ... ?

Yes as said, it's pre-draft. Thanks for spotting :)

> n the same document, when you write g(x) do you mean what in the Tinflex paper is given as Tc(x) ... ?

Yes `g(x)` was `\tilde{f}(x)`. Both functions are the result of the transformation. I updated the "Supplementary explanation doc" with a hopefully understandable explanation.
Don't hesitate to point out further problem in there.

joseph-wakeling-sociomantic: But `\tilde{f}(x)` is not the same as `T_c(x)` ... ?

wilzbach: > But \tilde{f}(x) is not the same as T_c(x) ... ?

As far as I understand is `T_c(x)` a family of two transformations
- `log(x)` for `c == 0`
- `sgn(c) * x^c` for `c != 0`

and `\tilde{f}{x} = T_c(f(x))`, so `f(x)` with the applied transformation.

It's a very confusing in the paper because they define a lot of special cases to avoid numerical errors in the common case.

wilzbach:
@9il @joseph-wakeling-sociomantic I have good news - I think we have the hat/squeeze generation working for all the test distributions from R. :tada: 

image

image

image

I also uploaded them to [Google Drive](https://drive.google.com/open?id=0BwdiZp7qSaBhQUNqSndVVjh4N00) for closer inspection.
They were generated with this command:

```
dub -c flex-plot -- -p=false
```

Note: `dist1_c` has currently an `undefined` interval (temporarily), I will look closer into this soon.
FYI the short flag `-p` is an abbreviation of `--plot_histogram`.

source/mir/random/flex/internal/area.d
joseph-wakeling-sociomantic, line 212: Hmmmm.  To me it just looks a bit weird to have one undocumented parameter in the middle, even though (as you say) it's pretty obvious what it is.  But anyway, not a big deal.

As for `isFloatingPoint` stuff, I would say it matters that this is verified at the outside user's point of entry into all this functionality, from there on in, it's fine to just assume.

source/mir/random/flex/internal/area.d
joseph-wakeling-sociomantic: Yea, matches what I understood.  I would suggest using `\tilde{f}` instead of `g` in your own writeup, just for clarity (I would define a custom LaTeX macro for it).

source/mir/random/flex/internal/area.d
```diff
@@ -0,0 +1,766 @@
...
+    {
+        z /= sh.a;
+        auto sgnc = copysign(S(1), iv.c);
+        if (!(sgnc * shL >= 0) || !(sgnc * shR >= 0))
+        {
+            area = S.max;
+        }
+        else if (iv.c == 1)
+        {
+            area = S(0.5) * sh.a * ivLength * (z + 2);
```
joseph-wakeling-sociomantic: Coming back to these `sh.a` contributions now that we are clear on the meaning of stuff in your writeup: what was bothering me about these components is that Table 2 explicitly gives the contribution as `f(x_0)` and not `\tilde{f}(x_0)`.  So the immediate relationship to `\alpha` wasn't obvious.

source/mir/random/flex/internal/area.d
wilzbach, line 212: > Hmmmm. To me it just looks a bit weird to have one undocumented parameter in the middle, even though (as you say) it's pretty obvious what it is.

Well my point was that we never document `S` in all other methods, so at least we are consistent. 

>  But anyway, not a big deal.

I don't mind adding it, @9il what do you say?

source/mir/random/flex/internal/area.d
wilzbach, line 87: > I repeat my earlier suggestion that the with statements here should be removed. What they save in typing, they remove in clarity of what's being altered (i.e. there is nothing in the code as written to suggest that squeeze and hat belong to iv).

ping @9il for your opinion

examples/flex_plot.d
```diff
@@ -316,11 +316,11 @@ void test_normal(S, F)(in ref F test)
 {
     import std.math : exp, log, PI, sqrt;
     S[] points = [-S.infinity, -1.5, 0, 1.5, S.infinity];
-    enum S sqrt2PI = sqrt(2 * PI);
+    enum S halfLog2PI = S(0.5) * log(2 * PI);
```
9il: why half?

wilzbach: because of the `0.5`, happy to change it for better naming suggestions ;-)

source/mir/random/flex/internal/calc.d
```diff
@@ -0,0 +1,140 @@
...
+        return tan(S(0.5) * (d + b));
+    }
+}
+
+/**
+Calculate the exponential mean between an Interval `iv`.
+
+ \int_l^r x * (h(x) - s(x)) dx
+ -----------------------------
+ \int^l^r      h(x) - s(x)  dx
```
joseph-wakeling-sociomantic, line 53: Just checking here: what's the difference between `\int_l^r` (in the numerator) and `\int^l^r` (in the denominator) ... ?

source/mir/random/flex/internal/calc.d
```diff
@@ -0,0 +1,140 @@
...
+        auto iv = Interval!S(-3, -1.36003, 1.5, 0, 0, 0, 0, 0, 0);
+        iv.hat = LF(0.159263, -1.36003, 1.26786),
+        iv.squeeze = LF(0.0200763, -3, 1.00667),
+        assert(expmean(iv).approxEqual(-1.90669));
+    }
+}
+
+/**
+Calculate the parameters for an interval.
+Given an interval, determine it's type and hat and squeeze function.
```
joseph-wakeling-sociomantic: `its`, not `it's`

joseph-wakeling-sociomantic: BTW I feel that your first line of doc here is misleading.  Starting with `Given an interval, determine its type, and its hat and squeeze functions` would be clearer.

I would also suggest including some description of the different interval types that can exist, just for clarity.

source/mir/random/flex/internal/calc.d
wilzbach, line 53: numerator: x*(h(x) - s(x))
denominator: h(x) - s(x)

joseph-wakeling-sociomantic, line 53: What I mean is, aren't your integral bounds specified wrongly for the denominator?

wilzbach:
this has been merged as part of #281 :tada:

PR #241, opened 2016-06-27
use in-place for area and apprMaxPoints in while loop

wilzbach:


wilzbach:
sorry for the noise - will directly push.

PR #248, opened 2016-07-04
use inverse method, z == 0 special cases, docs

wilzbach:
@9il this introduces two things
- `inverse` function for linear functions (used during the sampling step) - the idea is the code more legible and generic (as _any_ kind of linear function storage could be used)
- special case for `z == 0`

There are a couple of change mixed in this PR, I hope they don't disturb.


9il:
do not need z == 0 case

wilzbach:
> do not need z == 0 case

ok thanks.

PR #250, opened 2016-07-07
use local path for examples

wilzbach:
@9il I discovered that `dub` supports the `path` option to supply local dependencies -> let's clean our `dub.json`

### DIFF
--- a/examples/lda_hoffman_sparse.d
+++ b/examples/lda_hoffman_sparse.d
@@ -1,7 +1,7 @@
 #!/usr/bin/env dub
 /+ dub.sdl:
 name "lda_hoffman_sparse"
-dependency "mir" version=">0.15.1"
+dependency "mir" path=".."
 +/
 
 /++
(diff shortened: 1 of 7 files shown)


examples/lda_hoffman_sparse.d
```diff
@@ -26,7 +26,7 @@ void main(string[] args)
     if (args.length > 1)
         curFolder = args[1];
     else
-        curFolder = thisExePath.dirName;
+        curFolder = getcwd.buildPath("examples");
```
wilzbach, line 29: temporary build are done in `/tmp/...`, hence I guess `getcwd` is the best we could try?

wilzbach:
@9il It seems like I accidentally bumped into a `dub` bug - https://github.com/dlang/dub/issues/893

I added the following workaround to our builds:

```
dub fetch imageformats
dub build imageformats
```

It's not a perfect solution, but I hope that the `dub` bug gets resolved soon, s.t. we can remove this ;-)

9il:
fails

wilzbach:
> fails

passing (:

PR #251, opened 2016-07-07
clean dub.json from flex example and use single file syntax

wilzbach:
@9il our other examples also use SDLang and imho it's more readable for such a short list of examples.

See also #250 that does the cleanup for the other files

Plotting can be done with e.g. this command:

```
./examples/flex_plot.d -p=false -r=1.3 6
``
```

### DIFF
--- a/examples/flex_plot.d
+++ b/examples/flex_plot.d
@@ -1,6 +1,11 @@
-/**
-Simple plot
-*/
+#!/usr/bin/env dub
+/+ dub.sdl:
+name "flex_plot"
+dependency "mir" path=".."
+dependency "pyd" version="~>0.9.8"
+versions "Flex_logging"
+subConfigurations "pyd" "python35"
++/
 import mir.random.flex: FlexInterval;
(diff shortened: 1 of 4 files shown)


examples/flex_plot.d
```diff
@@ -1,6 +1,11 @@
-/**
-Simple plot
-*/
+#!/usr/bin/env dub
+/+ dub.sdl:
```
9il, line 2: use "dub.json" format (all sdl would be replaced)

wilzbach, line 2: but the `json` analog is a lot more verbose, uglier and harder to edit:

``` json
/+ dub.json:
{
    "name": "flex_plot",
    "dependencies": {
        "mir": {
            "path": ".."
        },
        "pyd": {
            "version": "~>0.9.8"
        }
    },
    "versions": ["Flex_logging"],
    "subConfigurations": {
        "pyd": "python35"
    }
}
++/
```

PR #259, opened 2016-07-18
Add discrete sampler based on Alias-method

### LINKED ISSUE #237: Discrete distribution sampling in O(1) using Alias or Marsaglia method

wilzbach:
Fixes #237.

It's based on 

> Vose, Michael D. "A linear algorithm for generating random numbers with a given distribution."
> IEEE Transactions on software engineering 17.9 (1991): 972-975.

which can be found [here](http://citeseerx.ist.psu.edu/viewdoc/download?doi=10.1.1.398.3339&rep=rep1&type=pdf).

tl;dr: Sampling in O(1) and initialization in O(n)

Ping @joseph-wakeling-sociomantic

(the PR renamed the current discrete sampling method to naive, however it's advantage is that no setup is required)


source/mir/random/discrete.d
```diff
@@ -11,6 +11,212 @@ module mir.random.discrete;
...
+*/
+Discrete!T discrete(T)(const(T)[] probs)
+{
+    return Discrete!T(probs);
+}
+
+///
+unittest
+{
+    auto probs = [0.1, 0.2, 0.5, 0.2];
```
joseph-wakeling-sociomantic, line 30: You may run into floating-point trouble with some compilers with probabilities based on the summation of `0.1` ;-)

It might be safer/saner to use 10, 20, 50, 20 as your proportions.

wilzbach, line 30: > You may run into floating-point trouble with some compilers with probabilities based on the summation of 0.1 ;-)

Yes even Vose91 anticipated that and thus inserted the cleanup loops at the end..

> It might be safer/saner to use 10, 20, 50, 20 as your proportions.

I don't know whether there is a huge difference, because 
a) if we multiply `50` with a large n we might run into the same troubles
b) this shifts the problem only a bit - e.g. `50.55` 
c) it's a less intuitive user API

source/mir/random/discrete.d
```diff
@@ -11,6 +11,212 @@ module mir.random.discrete;
...
+    */
+    this(const(T)[] probs)
+    {
+        debug
+        {
+            import std.algorithm.iteration : sum;
+            import std.math : approxEqual;
+            assert(probs.sum.approxEqual(1.0), "Sum of the probabilities must be 1");
+        }
+        initialize(probs);
```
wilzbach, line 78: As many other random algorithms also have a construction phase, it would be quite nice if we agree that the initialization is part of the struct initialization

source/mir/random/discrete.d
```diff
@@ -11,6 +11,212 @@ module mir.random.discrete;
...
+References:
+    Vose, Michael D. "A linear algorithm for generating random numbers with a given distribution."
+    IEEE Transactions on software engineering 17.9 (1991): 972-975.
+*/
+struct Discrete(T)
+    if (isNumeric!T)
+{
+
+    /// array with the original column value for a discrete value an
+    struct AltPair
```
9il: use `static`

source/mir/random/discrete.d
```diff
@@ -11,6 +11,212 @@ module mir.random.discrete;
...
+    Params:
+        probs = probabilities of the individual, discrete values
+
+    Complexity: O(n), where n is the number of discrete values
+    */
+    this(const(T)[] probs)
+    {
+        debug
+        {
+            import std.algorithm.iteration : sum;
```
9il: use fast summation from mir.sum

source/mir/random/discrete.d
```diff
@@ -11,6 +11,212 @@ module mir.random.discrete;
...
+    1/n - the average probability, so larger columns that are used to fill always
+    have enough capacity.
+    Afterwards in the sampling step such a column is picked by rolling a fair dice (1/n),
+    and whether the original or alias entry in the column should be used is picked
+    by tossing a coin with the probability equal to `p_column * n`.
+    A detailed proof can be found at [Vose91].
+    */
+    private void initialize(const(T)[] probs)
+    {
+        import std.container.slist : SList;
```
9il: can be done with single allocation in array. use `Mallocator` for allocation

source/mir/random/discrete.d
```diff
@@ -11,6 +11,212 @@ module mir.random.discrete;
...
+    /// samples a value from the discrete distribution using a custom random generator
+    size_t opCall(RNG)(ref RNG gen) const
+    {
+        import std.random : uniform;
+        import std.math : floor;
+
+        T u = uniform!("[)", T, T)(0, arr.length, gen);
+        size_t j = cast(size_t) floor(u);
+        auto vp = arr[j];
+        return (u - j <= vp.prob) ? j : vp.alt;
```
9il: just `arr[j].alt;`

9il: instead of `vp.alt;`

wilzbach:
> can be done with single allocation in array. use Mallocator for allocation

is the last commit how you imagined that?

source/mir/random/discrete.d
```diff
@@ -94,19 +94,36 @@ struct Discrete(T)
...
+
+            auto small = alloc.makeArray!ProbsIndexed(n); // columns that need to be filled
+            auto large = alloc.makeArray!ProbsIndexed(n); // used to fill columns
+            scope(exit)
+            {
+                alloc.dispose(small);
+                alloc.dispose(large);
+            }
+        }
+        else
```
9il: drop support for old DMD FE

source/mir/random/discrete.d
```diff
@@ -94,19 +94,36 @@ struct Discrete(T)
...
+
+        static if (__VERSION__ > 2069)
+        {
+            import std.experimental.allocator.mallocator : Mallocator;
+            import std.experimental.allocator : dispose, makeArray;
+            auto alloc = Mallocator.instance;
+
+            auto small = alloc.makeArray!ProbsIndexed(n); // columns that need to be filled
+            auto large = alloc.makeArray!ProbsIndexed(n); // used to fill columns
+            scope(exit)
```
9il: do not need `scope(exit)`. Do not use `scope` in low level math code

wilzbach:
> do not need scope(exit). Do not use scope in low level math code

Is https://github.com/libmir/mir/pull/259/commits/5290bbe27230a0cb08b87e0217bc9d3c0f0789c5 what you imagined?

9il:
yes

source/mir/random/discrete.d
```diff
@@ -11,6 +11,221 @@ module mir.random.discrete;
...
+{
+
+    /// array with the original column value for a discrete value an
+    static struct AltPair
+    {
+        T prob; /// probability p to select it by a coin toss, if this column is randomly picked
+        size_t alt; /// alternative value if coin toss at j fails
+    }
+
+    private AltPair[] arr;
```
wilzbach, line 61: should we provide the user a mean to use a custom Allocator for `arr`?

source/mir/random/discrete.d
```diff
@@ -11,6 +11,221 @@ module mir.random.discrete;
...
+            if (lg.prob < 1)
+                small[smCounter++] = lg;
+            else
+                large[lgCounter++] = lg;
+        }
+
+        // do cleanup
+        for (auto i = lgCounter; i > 0; i--)
+        {
+            arr[large[i - 1].index].prob = 1;
```
wilzbach, line 148: is there a smart way to avoid `i  - 1`, e.g. should I add an extra `if lgCounter > 0` on top and do `lgCounter - 1`?

source/mir/random/discrete.d
```diff
@@ -23,17 +238,17 @@ Params:
 Returns:
     A $(LREF Discrete) sampler that can be called to sample from the distribution.
 */
-Discrete!T discrete(T)(const(T)[] cdPoints)
+NaiveDiscrete!T naiveDiscrete(T)(const(T)[] cdPoints)
```
wilzbach, line 241: should I adapt `naiveDiscrete` to support an alternative probability-based interface, s.t. the API between both methods is similar?

source/mir/random/discrete.d
9il, line 148: `for (sizediff_t i = lgCounter - 1; i >= 0; i--)`

9il:
Ooops, merget unexpectably from phone

wilzbach:
I was about to write:

The performance of Alias with LDC looks good so far (see below).
@9il 
1) do we want to include such a benchmarking suite to mir?
2) Should I have a look at DMD's performance
3) Is the benchmarking approach correct? I used `extern` keywords in separate files for each method to avoid inlining. The files are linked together correctly but when templates are used the main file contains all the specialized templates.

```
>dmd
ds           : 631 ms, 57 μs, and 7 hnsecs
ds.naive: 597 ms and 498 μs
uniform : 398 ms, 581 μs, and 1 hnsec
>ldc
ds          : 324 ms, 150 μs, and 6 hnsecs
ds.naive: 411 ms, 963 μs, and 3 hnsecs
uniform : 315 ms, 922 μs, and 4 hnsecs
```

source/mir/random/discrete.d
```diff
@@ -11,6 +11,221 @@ module mir.random.discrete;
...
+        }
+
+        size_t smCounter, lgCounter;
+
+        import std.experimental.allocator.mallocator : Mallocator;
+        import std.experimental.allocator : dispose, makeArray;
+        auto alloc = Mallocator.instance;
+
+        auto small = alloc.makeArray!ProbsIndexed(n); // columns that need to be filled
+        auto large = alloc.makeArray!ProbsIndexed(n); // used to fill columns
```
9il, line 113: One array should be enough. It would be a little bit more cache friendly, and require only one system call. The idea is use the end of array for reversed stack.

9il:
> I was about to write:
> 
> The performance of Alias with LDC looks good so far (see below).
> @9il 
> 1) do we want to include such a benchmarking suite to mir?
> 2) Should I have a look at DMD's performance
> 3) Is the benchmarking approach correct? I used extern keywords in separate files for each method to avoid inlining. The files are linked together correctly but when templates are used the main file contains all the specialized templates.
> 
> > dmd
> > ds           : 631 ms, 57 μs, and 7 hnsecs
> > ds.naive: 597 ms and 498 μs
> > uniform : 398 ms, 581 μs, and 1 hnsec
> > ldc
> > ds          : 324 ms, 150 μs, and 6 hnsecs
> > ds.naive: 411 ms, 963 μs, and 3 hnsecs
> > uniform : 315 ms, 922 μs, and 4 hnsecs

We need benchmarks but for algorithms with the same complexity. Please remove this test. You can add more information about algorithms complexity and references if you want

One informative chart is much better then code.

9il:
so please remove current test. We would need a lot of benchmarks letter. The should be simple for a user and please no make files. Inlining is OK

PR #260, opened 2016-07-18
fix travis for imageformats

wilzbach:
@9il I don't know what happened, but this doesn't look good:

image

Let's see whether this quick fix resolves the problem.


wilzbach:
seems like Travis healed itself :)

PR #261, opened 2016-07-19
mir.random.nonuniform: Add Ziggurat method for Normal & Exponential

wilzbach:
Adds the Ziggurat sampling algorithm for Normal & Exponential distribution.

1) [Paper](http://www.jstatsoft.org/v05/i08/paper)

>   Marsaglia, George, and Wai Wan Tsang. "The ziggurat method for generating random variables."
>     Journal of statistical software 5.8 (2000): 1-7.

2) Thoughts
-  [ ] there are two other important papers
  -  [design flaws of the Ziggurat algorithm](http://arxiv.org/pdf/math/0603058) - the main critic is that the same variable is used to pick the block & value
  -  [An Improved Ziggurat Method to Generate Normal Random Samples](http://www.doornik.com/research/ziggurat.pdf)
- [ ] it doesn't look perfect yet, the mean is `-0.02` :/
- [ ] Ziggurat works for all distributions that can be reduced to a all monotone decreasing distributions (-> will assemble a list tomorrow)

3) Distribution plots (`dub ./examples/nonuniform_plot.d`):

image

image

Ping @joseph-wakeling-sociomantic @9il

### DIFF
--- a/source/mir/random/nonuniform.d
+++ b/source/mir/random/nonuniform.d
@@ -50,7 +50,7 @@ auto normal(T, UIntType = uint)()
         }
     };
 
-    return Ziggurat!(T, fallback, UIntType, true)(pdf, invPdf, 128, rightEnd, T(9.91256303526217e-3));
+    return Ziggurat!(T, fallback, UIntType, 128, true)(pdf, invPdf, rightEnd, T(9.91256303526217e-3));
 }
 
 /**
@@ -80,7 +80,7 @@ auto exponential(T, UIntType = uint)()
         }
     };
 
-    return Ziggurat!(T, fallback, UIntType, false)(pdf, invPdf, 256, T(7.697117470131487), T(3.949659822581572e-3));
+    return Ziggurat!(T, fallback, UIntType, 256, false)(pdf, invPdf, T(7.697117470131487), T(3.949659822581572e-3));
 }
 
 /**
@@ -92,7 +92,7 @@ References:
     Marsaglia, George, and Wai Wan Tsang. "The ziggurat method for generating random variables."
     Journal of statistical software 5.8 (2000): 1-7.
 */
-struct Ziggurat(T, string _fallback, UIntType, bool bothSides)
+struct Ziggurat(T, string _fallback, UIntType, size_t numberOfBlocks, bool bothSides)
     if (isNumeric!T)
 {
 
@@ -104,16 +104,14 @@ private:
     T function(T x) invPdf;
 
     /// precalculate difference x_i / x_{i+1}
-    T[] xDiv;
+    T[numberOfBlocks] xDiv;
     /// precalculate scaling to R.max for x_i
-    T[] xScaled;
+    T[numberOfBlocks] xScaled;
     /// precalculate pdf value for x_i
-    T[] fs;
+    T[numberOfBlocks] fs;
 
-    /// number of blocks
-    size_t k;
     // mask to use to get the first log_2 k bits (=k - 1)
-    size_t kMask;
+    enum size_t kMask = numberOfBlocks - 1;
 
     /// left-point of the right-most block
     T rightEnd;
@@ -122,39 +120,33 @@ private:
     T blockArea;
 
 public:
-    this(T function(T x) pdf, T function(T x) invPdf, size_t k, T rightEnd,
-         T blockArea)
+    this(T function(T x) pdf, T function(T x) invPdf, T rightEnd, T blockArea)
     {
         this.pdf = pdf;
         this.invPdf = invPdf;
-        this.k = k;
         this.rightEnd = rightEnd;
         this.blockArea = blockArea;
 
         // scale factor to the range of UIntType
         T maxT = bothSides ? UIntType.max / 2 : UIntType.max;
-        kMask = k - 1;
 
         auto xn = rightEnd; // Next x_{i+1}
         auto xc = xn; // Current x_i
         auto fn = pdf(xn);
 
         // k_i = floor(2^32 (x_{i-1} / x_i))
-        xDiv    = new T[k];
         xDiv[0] = (xn * fn / blockArea) * maxT;
         xDiv[1] = 0;
 
         // w_i = 0.5^32 * x_i
-        xScaled        = new T[k];
         xScaled[0]     = (blockArea / fn) / maxT;
         xScaled[$ - 1] = xn / maxT;
 
         // f_i = f(x_i)
-        fs        = new T[k];
         fs[0]     = T(1);
         fs[$ - 1] = fn;
 
-        for (auto i = k - 2; i >= 1; i--)
+        for (auto i = xScaled.length - 2; i >= 1; i--)
         {
             xn = invPdf(blockArea / xn + fn);
             xDiv[i + 1] = (xn / xc) * maxT;
@@ -173,6 +165,7 @@ public:
 
     /// samples a value from the discrete distribution using a custom random generator
     T opCall(RNG)(ref RNG gen) const
+        if (typeof(gen.front).sizeof == UIntType.sizeof)
     {
         import std.random : uniform;
         import std.traits : Signed;
--- a/source/mir/random/nonuniform.d
+++ b/source/mir/random/nonuniform.d
@@ -120,6 +120,13 @@ private:
     T blockArea;
 
 public:
+    /**
+    Params:
+        pdf = probability density function
+        invPdf = inverse probability density function
+        rightEnd = left point of the right-most block
+        blockArea =  area of every column
+    */
     this(T function(T x) pdf, T function(T x) invPdf, T rightEnd, T blockArea)
     {
         this.pdf = pdf;
--- a/source/mir/random/nonuniform.d
+++ b/source/mir/random/nonuniform.d
@@ -16,12 +16,12 @@ Setup a _normal distribution sampler.
 
 Params:
     T = floating-point type that should be sampled
-    R = unsigned UIntType of the random generator
+    UIntType = unsigned UIntType of the random generator
 
 Returns:
     A $(LREF Ziggurat) sampler that can be called to sample from the distribution.
 */
-auto normal(T, R = uint)()
+auto normal(T, UIntType = uint)()
 {
     import mir.internal.math : exp, log, sqrt;
 
@@ -50,7 +50,7 @@ auto normal(T, R = uint)()
         }
     };
 
-    return Ziggurat!(T, fallback, R, true)(pdf, invPdf, 128, rightEnd, T(9.91256303526217e-3));
+    return Ziggurat!(T, fallback, UIntType, true)(pdf, invPdf, 128, rightEnd, T(9.91256303526217e-3));
 }
 
 /**
@@ -63,7 +63,7 @@ Params:
 Returns:
     A $(LREF Ziggurat) sampler that can be called to sample from the distribution.
 */
-auto exponential(T, R = uint)()
+auto exponential(T, UIntType = uint)()
 {
     import mir.internal.math : exp, log;
 
@@ -80,7 +80,7 @@ auto exponential(T, R = uint)()
         }
     };
 
-    return Ziggurat!(T, fallback, R, false)(pdf, invPdf, 256, T(7.697117470131487), T(3.949659822581572e-3));
+    return Ziggurat!(T, fallback, UIntType, false)(pdf, invPdf, 256, T(7.697117470131487), T(3.949659822581572e-3));
 }
 
 /**
@@ -92,7 +92,7 @@ References:
     Marsaglia, George, and Wai Wan Tsang. "The ziggurat method for generating random variables."
     Journal of statistical software 5.8 (2000): 1-7.
 */
-struct Ziggurat(T, string _fallback, R = uint, bool bothSides)
+struct Ziggurat(T, string _fallback, UIntType, bool bothSides)
     if (isNumeric!T)
 {
 
@@ -132,7 +132,7 @@ public:
         this.blockArea = blockArea;
 
         // scale factor to the range of UIntType
-        T maxT = bothSides ? R.max / 2 : R.max;
+        T maxT = bothSides ? UIntType.max / 2 : UIntType.max;
         kMask = k - 1;
 
         auto xn = rightEnd; // Next x_{i+1}
@@ -177,7 +177,7 @@ public:
         import std.random : uniform;
         import std.traits : Signed;
         import std.range : ElementType;
-        import std.mir.internal : fabs;
+        import std.math : abs;
 
         // TODO: option for inlining
         for (;;)
@@ -196,7 +196,7 @@ public:
             // U * x_i < x_{i+1}
             static if (bothSides)
             {
-                if (fabs(u) < xDiv[i])
+                if (abs(u) < xDiv[i])
                     return u * xScaled[i];
             }
             else
--- a/source/mir/random/nonuniform.d
+++ b/source/mir/random/nonuniform.d
@@ -45,7 +45,7 @@ auto normal(T, UIntType = uint)()
                 u = uniform!("[]", T, T)(0, 1, gen);
                 y = -log(u);
             }
-            while(y + y < x * x);
+            while (y + y < x * x);
             return isPositive ? rightEnd + x : -rightEnd - x;
         }
     };
--- a/source/mir/random/nonuniform.d
+++ b/source/mir/random/nonuniform.d
@@ -1,0 +1,246 @@
+/**
+Non-uniform distributions.
+
+License: $(LINK2 http://boost.org/LICENSE_1_0.txt, Boost License 1.0).
+
+Authors: Sebastian Wilzbach
+*/
+
+module mir.random.nonuniform;
+
+import std.traits : isNumeric;
+import std.stdio;
+
+/**
+Setup a _normal distribution sampler.
+
+Params:
+    T = floating-point type that should be sampled
+    R = unsigned UIntType of the random generator
+
+Returns:
+    A $(LREF Ziggurat) sampler that can be called to sample from the distribution.
+*/
+auto normal(T, R = uint)()
+{
+    import mir.internal.math : exp, log, sqrt;
+
+    auto pdf    = (T x) => cast(T) exp(T(-0.5) * x * x);
+    auto invPdf = (T x) => cast(T) sqrt(T(-2) * log(x));
+
+    // values from [Marsaglia00]
+    T rightEnd = 3.442619855899;
+
+    /// generate x for the last block (aka tail-block)
+    enum fallback = q{
+        T fallback(RNG)(bool isPositive, ref RNG gen) const
+        {
+            import std.random : uniform;
+            import std.math : log;
+            T x, y, u;
+            do
+            {
+                u = uniform!("[]", T, T)(0, 1, gen);
+                x = -log(u) / fs[$ - 1];
+                u = uniform!("[]", T, T)(0, 1, gen);
+                y = -log(u);
+            }
+            while(y + y < x * x);
+            return isPositive ? rightEnd + x : -rightEnd - x;
+        }
+    };
+
+    return Ziggurat!(T, fallback, R, true)(pdf, invPdf, 128, rightEnd, T(9.91256303526217e-3));
+}
+
+/**
+Setup an _exponential distribution sampler.
+
+Params:
+    T = floating-point type that should be sampled
+    R = unsigned UIntType of the random generator
+
+Returns:
+    A $(LREF Ziggurat) sampler that can be called to sample from the distribution.
+*/
+auto exponential(T, R = uint)()
+{
+    import mir.internal.math : exp, log;
+
+    auto pdf    = (T x) => cast(T) exp(-x);
+    auto invPdf = (T x) => cast(T) -log(x);
+
+    // values from [Marsaglia00]
+    enum fallback = q{
+        T fallback(RNG)(ref RNG gen) const
+        {
+            import std.random : uniform;
+            auto u = uniform!("[]", T, T)(0, 1, gen);
+            return 7.69711 - u;
+        }
+    };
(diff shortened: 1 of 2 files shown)


source/mir/random/nonuniform.d
```diff
@@ -0,0 +1,246 @@
...
+                x = -log(u) / fs[$ - 1];
+                u = uniform!("[]", T, T)(0, 1, gen);
+                y = -log(u);
+            }
+            while(y + y < x * x);
+            return isPositive ? rightEnd + x : -rightEnd - x;
+        }
+    };
+
+    return Ziggurat!(T, fallback, R, true)(pdf, invPdf, 128, rightEnd, T(9.91256303526217e-3));
```
wilzbach: I actually would like to run the initialization in CTFE as it will never change, but `exp`  uses inline assembler which isn't supported in CTFE :/
Has anyone an idea?

source/mir/random/nonuniform.d
```diff
@@ -0,0 +1,246 @@
...
+    /// constants don't matter
+    T function(T x) pdf;
+    T function(T x) invPdf;
+
+    /// precalculate difference x_i / x_{i+1}
+    T[] xDiv;
+    /// precalculate scaling to R.max for x_i
+    T[] xScaled;
+    /// precalculate pdf value for x_i
+    T[] fs;
```
wilzbach: "An Improved Ziggurat Method to Generate Normal Random Samples" postulates that saving the `fs` isn't necessary to which I couldn't follow yet.

> DRanU() returns a uniform random number, U (0, 1), and IRanU() returns 32-bit
> unsigned random integer. DRanNormalTail is implemented as a separate function:
> it gets called only rarely, so that efficiency does not matter. For the same reason, it
> is not necessary to avoid a call to exp() when checking for the wedges (this could
> be achieved by precomputing the function values f (xi). (page 7)

Imho saving doesn't waste much space and saves time

source/mir/random/nonuniform.d
```diff
@@ -0,0 +1,246 @@
...
+
+private:
+
+    /// monotone decreasing probability density function
+    /// constants don't matter
+    T function(T x) pdf;
+    T function(T x) invPdf;
+
+    /// precalculate difference x_i / x_{i+1}
+    T[] xDiv;
```
wilzbach: As the array size is known at compile-time, should we use `T[k]`?

source/mir/random/nonuniform.d
joseph-wakeling-sociomantic: Well, start by filing an issue against phobos asking for a CTFE'able `exp`.

BTW where on earth does this magic constant `9.91256...e-3` come from?  I would suggest making it a named manifest constant.

wilzbach: > Well, start by filing an issue against phobos asking for a CTFE'able exp.

Ok thanks - [done](https://issues.dlang.org/show_bug.cgi?id=16296). I will test whether copying the non-inline version from Phobos works.

> BTW where on earth does this magic constant 9.91256...e-3 come from? I would suggest making it a named manifest constant.

It also comes from  [Marsaglia00] - there it's called `v`.
It's the area of every block and thus depends on `k` and the distribution.

-> I will declare it more explicitly.

source/mir/random/nonuniform.d
```diff
@@ -0,0 +1,246 @@
...
+/**
+Ziggurat algorithm for generating a random variable from a montone, decreasing density.
+
+Complexity: O_avg: O(1)
+
+References:
+    Marsaglia, George, and Wai Wan Tsang. "The ziggurat method for generating random variables."
+    Journal of statistical software 5.8 (2000): 1-7.
+*/
+struct Ziggurat(T, string _fallback, R = uint, bool bothSides)
```
joseph-wakeling-sociomantic: Instead of `R` I would use `UIntType` (it matches the typical template-parameter name in both phobos `std.random` and C++11 `<random>` for the word-type of the uniform RNG).

However, in this case, is there any possibility to avoid needing the word-type to be known at compile time?

source/mir/random/nonuniform.d
```diff
@@ -0,0 +1,246 @@
...
+
+    /// samples a value from the discrete distribution
+    T opCall() const
+    {
+        import std.random : rndGen;
+        return opCall(rndGen);
+    }
+
+    /// samples a value from the discrete distribution using a custom random generator
+    T opCall(RNG)(ref RNG gen) const
```
joseph-wakeling-sociomantic, line 174: Assuming you do need to know the word-type (your template parameter `R`) at compile time, you might want to validate that `typeof(gen.front)` matches it.

Note that in principle at least it's possible to handle different `R` and generator return type: if R's size is a multiple of the generator's word size, then you can populate it by several calls to the generator; conversely, if `R` is smaller than the generator's word size, you can use a single RNG variate to provide several of the needed values.  But this may be adding excessive complexity.

source/mir/random/nonuniform.d
joseph-wakeling-sociomantic: Also, note that while a dependency on the word size is potentially more flexible, it's worth considering also just templating the ziggurat on the actual RNG type.

source/mir/random/nonuniform.d
```diff
@@ -0,0 +1,246 @@
...
+
+Returns:
+    A $(LREF Ziggurat) sampler that can be called to sample from the distribution.
+*/
+auto exponential(T, R = uint)()
+{
+    import mir.internal.math : exp, log;
+
+    auto pdf    = (T x) => cast(T) exp(-x);
+    auto invPdf = (T x) => cast(T) -log(x);
```
joseph-wakeling-sociomantic, line 71: I'm not sure I like the explicit casts.  Surely it's possible to get the same result here without them?

source/mir/random/nonuniform.d
```diff
@@ -0,0 +1,246 @@
...
+    enum fallback = q{
+        T fallback(RNG)(ref RNG gen) const
+        {
+            import std.random : uniform;
+            auto u = uniform!("[]", T, T)(0, 1, gen);
+            return 7.69711 - u;
+        }
+    };
+
+    return Ziggurat!(T, fallback, R, false)(pdf, invPdf, 256, T(7.697117470131487), T(3.949659822581572e-3));
```
joseph-wakeling-sociomantic: While calculations using `pdf` and `invPdf` can only be done at runtime for now (because of the `exp` implementation issues), is there any reason why the actual lambdas can't be provided as template parameters?  It would make for a more logical design, I think (and also be future-proof against the point when you get a CTFE'able `exp`).

joseph-wakeling-sociomantic: I would also suggest defining explicit templates for `Normal(T)`, `Exponential(T)` etc. to wrap these `Ziggurat` instantiations.

joseph-wakeling-sociomantic: An alternative would be to define actual `Normal(T)` and `Exponential(T)` wrapper structs that just use a `Ziggurat` instance internally; that might also give you a more future-proof API should you ever wish to rework the internals.

That might be better, because it'll probably be easier for users to debug stuff if they see a type called `Normal!double` instead of `Ziggurat!(LOTS, OF, DIFFERENT, STUFF)`.

wilzbach: I put everything as template argument that is absolutely needed at compile time. 
Does it make a huge difference? - I thought in the end we can just use `enum z = Ziggurat!..`

joseph-wakeling-sociomantic: It makes less of a difference if the `Ziggurat` instance is wrapped away inside a type that's explicitly `Normal` or `Exponential` or whatever.  But if you're set on returning a raw `Ziggurat` instantiation from your `normal` and `exponential` factory functions, it might be useful to have the PDF and inverse PDF clearly there in the template parameters.

source/mir/random/nonuniform.d
joseph-wakeling-sociomantic: Don't see why not, but right now `k` is _not_ provided at compile time AFAICS?

source/mir/random/nonuniform.d
wilzbach: > An alternative would be to define actual Normal(T) and Exponential(T) wrapper structs that just use a Ziggurat instance internally; that might also give you a more future-proof API should you ever wish to rework the internals.

I really like the idea of having a broad `Normal(T, UIntType)` API, but with my current attempt (see below - for some reasons the PR didn't go through yesterday) it is yet another function call.

https://github.com/wilzbach/mir/pull/6/commits/4c2f4bd5f7d2b0e0b0ebdb8dcd2255e52cf2357
9

Should I build a Ziggurat struct with mixins?

source/mir/random/nonuniform.d
wilzbach, line 71: (Idk why my comment wasn't saved). The problem is that `exp` only accepts and returns `real`.

source/mir/random/nonuniform.d
wilzbach, line 174: > Assuming you do need to know the word-type (your template parameter R) at compile time, you might want to validate that typeof(gen.front) matches it.

Nice idea (done).

> Note that in principle at least it's possible to handle different R and generator return type: if R's size is a multiple of the generator's word size, then you can populate it by several calls to the generator; conversely, if R is smaller than the generator's word size, you can use a single RNG variate to provide several of the needed values. But this may be adding excessive complexity.

Good point - how common is it to have something different than `uint` or `ulong`?

joseph-wakeling-sociomantic, line 174: > Good point - how common is it to have something different than uint or ulong?

Not common, I think, at least not these days.

source/mir/random/nonuniform.d
wilzbach: > However, in this case, is there any possibility to avoid needing the word-type to be known at compile time?

Well you already proposed it.

> Also, note that while a dependency on the word size is potentially more flexible, it's worth considering also just templating the ziggurat on the actual RNG type.

Hmm while I understand the motivation of providing a simple API for the user, this would create even more template bloat (for every RNG, RNG template in Ziggurat template) and limit the API more than needed. Hence I think making `opCall` generic depending on the RNG is probably the better way to go.

source/mir/random/nonuniform.d
wilzbach: > Don't see why not, but right now k is not provided at compile time AFAICS?

Well it's not strictly needed at compile-time (in comparison to the other compile-time parameters), but I don't see any use case where one wants to be able to choose the block size at runtime as the `rightEnd` and `averageArea` need to be precomputed by hand anyways.

wilzbach:
(will post more summaries here soon, here's a brief overview)

### [Design flaws of the Ziggurat algorithm](http://arxiv.org/pdf/math/0603058)

two main critics:
- same variable is used to pick the block & value (last 7 or 8 bits). However `2^50` values are needed to detect this
  - `SHR3` not uniform (shouldn't affect us)

### [An Improved Ziggurat Method to Generate Normal Random Samples](http://www.doornik.com/research/ziggurat.pdf)
- use `double`
- two random variables (one separate 7bit one to pick the block) -> optimization uses 1 + 1/4 random variables
- no precomputation of `f(x_i)` 
- idea: use two random integers for higher-precisions values

wilzbach:
btw an interesting overview paper is [Gaussian Random Number Generators](http://www.doc.ic.ac.uk/~wl/papers/07/csur07dt.pdf) by Thomas et. al.
Summary: Wallace-method is the fastest, but doesn't provide good statistical quality. Ziggurat was the second fastest method among the huge benchmark while passing the chi-squared and achieving good scores at the high sigma test (see table 3 and 4).

source/mir/random/nonuniform.d
```diff
@@ -0,0 +1,246 @@
+/**
+Non-uniform distributions.
+
+License: $(LINK2 http://boost.org/LICENSE_1_0.txt, Boost License 1.0).
+
+Authors: Sebastian Wilzbach
+*/
+
+module mir.random.nonuniform;
```
joseph-wakeling-sociomantic, line 9: Minor, but didn't we agree to create a `mir.random.distribution` package to contain everything ... ?

wilzbach, line 9: See: https://github.com/libmir/mir/issues/262

source/mir/random/nonuniform.d
```diff
@@ -0,0 +1,246 @@
...
+        {
+            static if (bothSides)
+                Signed!(ElementType!RNG) u = gen.front;
+            else
+                auto u = gen.front;
+
+            gen.popFront();
+
+            // TODO: this is a bit biased
+            size_t i = u & kMask;
```
joseph-wakeling-sociomantic, line 193: Given your `TODO` here, I assume this is the first place you're looking to understand why the plotted distributions don't look quite right .... ?

wilzbach, line 193: Yes resolving whether we can reuse the random bits or need to generate another 1/4 random variable (one of the optimizations described in later papers), shouldn't matter so much :/

source/mir/random/nonuniform.d
```diff
@@ -0,0 +1,246 @@
...
+    /// generate x for the last block (aka tail-block)
+    enum fallback = q{
+        T fallback(RNG)(bool isPositive, ref RNG gen) const
+        {
+            import std.random : uniform;
+            import std.math : log;
+            T x, y, u;
+            do
+            {
+                u = uniform!("[]", T, T)(0, 1, gen);
```
joseph-wakeling-sociomantic, line 43: Marsaglia & Tang talk about `UNI` (in their paper) being a generator of "uniform (0, 1) variates".  That would suggest over the open rather than closed interval, i.e. `uniform!"()"` rather than `uniform!"[]"`.  Does that make a difference to your distribution results?

source/mir/random/nonuniform.d
```diff
@@ -0,0 +1,246 @@
...
+            static if (bothSides)
+                Signed!(ElementType!RNG) u = gen.front;
+            else
+                auto u = gen.front;
+
+            gen.popFront();
+
+            // TODO: this is a bit biased
+            size_t i = u & kMask;
+            //size_t i = uniform!("[)", size_t, size_t)(0, kMask, gen);
```
joseph-wakeling-sociomantic, line 194: Minor: note that `uniform!T` should give uniform distribution across all possible values of an integral type `T`.  But in this case arguably unnecessary.

source/mir/random/nonuniform.d
```diff
@@ -0,0 +1,246 @@
...
+
+Returns:
+    A $(LREF Ziggurat) sampler that can be called to sample from the distribution.
+*/
+auto normal(T, UIntType = uint)()
+{
+    import mir.internal.math : exp, log, sqrt;
+
+    auto pdf    = (T x) => cast(T) exp(T(-0.5) * x * x);
+    auto invPdf = (T x) => cast(T) sqrt(T(-2) * log(x));
```
joseph-wakeling-sociomantic, line 29: Aren't these definitions missing the mean and variance parameters of the normal distribution?  I recognize that one can generate any normal distribution once one has variates from N(0, 1), but surely that should be baked in to the implementation?

Also: even allowing for N(0, 1), isn't the PDF missing the divisor by `\sqrt{2 * \pi}` ... ?  & presumably this means the inverse PDF is also missing something correspondingly?

source/mir/random/nonuniform.d
joseph-wakeling-sociomantic, line 71: Just as the `normal` implementation looks to be missing mean + variance, here the exponential implementation is missing its `\lambda` control parameter.

9il:
Does Ziggurat method yield better results for normal distribution then other methods in Atmosphere or dstats?

wilzbach:
> Does Ziggurat method yield better results for normal distribution then other methods in Atmosphere or dstats?

What's the best way to compare?

wilzbach:
@joseph-wakeling-sociomantic

NormalDist CDF looks good from what I can judge

image

Exp doesn't

image

9il:
> Does Ziggurat method yield better results for normal distribution then other methods in Atmosphere or dstats?
> What's the best way to compare?

If you don't know if Ziggurat yield better result, then we do not need Ziggurat

9il:
> What's the best way to compare?

There are no the best way. The best option is to read a couple of articles. We need to understand need we Ziggurat or not first, before spend time on it

joseph-wakeling-sociomantic:
> If you don't know if Ziggurat yield better result, then we do not need Ziggurat

The existing academic literature would suggest that Ziggurat is a very effective method; the paper by Thomas et al. offers a variety of tests of statistical quality that could be used, IIRC.

Question is, given the timelines, is it worth pushing on with Ziggurat, or would it be better to implement more basic implementations of the various distributions, and return to Ziggurat as a longer-term work?

9il:
> Question is, given the timelines, is it worth pushing on with Ziggurat, or would it be better to implement more basic implementations of the various distributions, and return to Ziggurat as a longer-term work?

First question is Ziggurat better for basic distributions than basic implementations of them? If there no strong `Yes`, then we the next after Tinflex is basic implementations.

wilzbach:
> The existing academic literature would suggest that Ziggurat is a very effective method; the paper by Thomas et al. offers a variety of tests of statistical quality that could be used, IIRC.

Yes sorry I should have been more precise. @9il what would be needed to convince you that Ziggurat is better than the algorithms in Atmosphere or dstats?
Is a X^2 test (that's what they use to evaluate the goodness of the fit) ok or should I also do the high-sigma test (more complex)?

> Question is, given the timelines, is it worth pushing on with Ziggurat, or would it be better to implement more basic implementations of the various distributions, and return to Ziggurat as a longer-term work?

We have time at least time until October. I would prefer to go with the "better" algorithm and tune it. We already have the basics implementations in dstats against which we can benchmark and compare.

wilzbach:
> First question is Ziggurat better for basic distributions than basic implementations of them? If there no strong Yes, then we the next after Tinflex is basic implementations.

Quote from my comment above:

> btw an interesting overview paper is Gaussian Random Number Generators by Thomas et. al.
> Summary: Wallace-method is the fastest, but doesn't provide good statistical quality. Ziggurat was the second fastest method among the huge benchmark while passing the chi-squared and achieving good scores at the high sigma test (see table 3 and 4).

9il:
> btw an interesting overview paper is Gaussian Random Number Generators by Thomas et. al.
> Summary: Wallace-method is the fastest, but doesn't provide good statistical quality. Ziggurat was the second fastest method among the huge benchmark while passing the chi-squared and achieving good scores at the high sigma test (see table 3 and 4).

Wallace-method is not specialised for Normal if I am not wrong

wilzbach:
> Wallace-method is not specialised for Normal if I am not wrong

Afaik Ziggurat is neither. In the literature it's just commonly only used for Normal and exponential distributions, however it's a general method that works for all monotone decreasing distributions (or if their symmetric half is monotone decreasing)

joseph-wakeling-sociomantic:
> First question is Ziggurat better for basic distributions than basic implementations of them?

I would say, "Yes, but."  The "but" is because Ziggurat is more complicated to implement correctly (as we're learning).  So, in terms of the current project, I would say there's a tradeoff between focusing on Ziggurat correct, versus getting a good variety of basic distributions in place with simpler (but more limited) algorithms.

joseph-wakeling-sociomantic:
> however it's a general method that works for all monotone decreasing distributions (or if their symmetric half is monotone decreasing

Yes, exactly.

WebDrake:
@wilzbach:

> We have time at least time until October.

The emails I'm getting from GSoC suggest that we're supposed to be finished up by the end of August, with 23 August as your own deadline for finalizing code and 29 August as the deadline for Ilya and me to submit our final evaluation report?
https://developers.google.com/open-source/gsoc/timeline

wilzbach:
> The emails I'm getting from GSoC suggest that we're supposed to be finished up by the end of August, with 23 August as your own deadline for finalizing code and 29 August as the deadline for Ilya and me to submit our final evaluation report?
> https://developers.google.com/open-source/gsoc/timeline

Yep, but that doesn't stop me to continue to work (I know that I wasted quite a lot of time)
My submission will only include the (Tin)flex algorithm, but I am still onto the mission to write (the building blocks for) a new & fast std.random for D. Hence I suggested to do it properly and as the benchmark in #286 suggests it's worth it.

WebDrake:
> Yep, but that doesn't stop me to continue to work (I know that I wasted quite a lot of time)

It's great that you want to keep working, but I was concerned about the expectations raised in the description of your GSoC project and what the people responsible might expect to see (which is why earlier I raised the possibility of doing some basic implementations of a variety of non-uniform distributions).

@9il you're the primary mentor here, so what are your thoughts?

9il:
> @9il you're the primary mentor here, so what are your thoughts?

We already have 2 general purpose Discrete RNG realizations and Tinflex will be ready soon.

Tinflex is hard numeric project without obvious workforce requirements. R version contains a lot of numeric bugs, many of them are fixed in the this project. We can stamp / copy-past boost rng, and this is not problem. But copy-pasting is not related to the proper RNG numbers. First, we need proper shell over std.random and fixed `uniform` generators. @wilzbach expected that he would add them during GSoC, in the same I didn't expect it.

If @wilzbach implemented a variety of non-uniform distributions but not Tinflex, then I would not be able to consider this GSoC project as finished.

WebDrake:
> If @wilzbach implemented a variety of non-uniform distributions but not Tinflex, then I would not be able to consider this GSoC project as finished.

I agree with that; I'm asking whether we should implement a variety of non-uniform distributions instead of (short term) focusing on Ziggurat.  Tinflex obviously takes primacy as it is the most significant part of the promised work.

9il:
> I agree with that; I'm asking whether we should implement a variety of non-uniform distributions instead of (short term) focusing on Ziggurat. Tinflex obviously takes primacy as it is the most significant part of the promised work.

Yes, I prefer to add more a variety of non-uniform distributions instead of Ziggurat

9il:
Please reopen for mir-random

PR #263, opened 2016-07-19
Simplify Tinflex API

wilzbach:
I think it's a reasonable assumption that all functions have the same type?

(see also #240)

### DIFF
--- a/source/mir/random/flex/package.d
+++ b/source/mir/random/flex/package.d
@@ -6,9 +6,9 @@ License: $(LINK2 http://boost.org/LICENSE_1_0.txt, Boost License 1.0).
 Authors: Sebastian Wilzbach, Ilya Yaroshenko
 
 The Transformed Density Rejection with Inflection Points (Flex) algorithm
-can sample from arbitrary distributions given its density function f, its
-first two derivatives and a partitioning into intervals with at most one inflection
-point.
+can sample from arbitrary distributions given (1) its log-density function f,
+(2) its first two derivatives and (3) a partitioning into intervals
+with at most one inflection point.
 
 These can be easily found by plotting `f''`.
 $(B Inflection point) can be identified by observing at which points `f''` is 0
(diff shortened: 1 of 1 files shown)


9il:
please keep old api

wilzbach:
> please keep old api

ok, but you explicitly instructed me on #240 for this API and it made sense to me as the pdf is used for sampling.

wilzbach:
> please keep old api

rebased.

source/mir/random/flex/package.d
```diff
@@ -110,6 +111,10 @@ Returns:
 auto flex(F0, F1, F2, S)
```
9il: `S` should be the first argument.

source/mir/random/flex/package.d
```diff
@@ -110,6 +111,10 @@ Returns:
...
                (in F0 f0, in F1 f1, in F2 f2,
                 S c, S[] points, S rho = 1.1)
+    if (isFloatingPoint!(ReturnType!F0) &&
```
9il: use `allSatisfy`

9il: BTW, function can return integer

source/mir/random/flex/package.d
```diff
@@ -118,18 +123,52 @@ auto flex(F0, F1, F2, S)
 }
 
 /// ditto
+auto flex(Pdf, F0, F1, F2, S)
```
9il: Ditto

source/mir/random/flex/package.d
9il: Just remove this constraints. They would not work for templated opCall for example

source/mir/random/flex/package.d
```diff
@@ -118,18 +123,52 @@ auto flex(F0, F1, F2, S)
...
+               (in Pdf pdf, in F0 f0, in F1 f1, in F2 f2,
+                S c, S[] points, S rho = 1.1)
+    if (isFloatingPoint!(ReturnType!Pdf) &&
```
9il: remove this constraints. Save only `isFloatingPoint!S`

wilzbach:
rebased API

source/mir/random/flex/package.d
```diff
@@ -118,26 +120,49 @@ auto flex(F0, F1, F2, S)
...
 }
 
 /// ditto
-auto flex(F0, S)(in F0 f0, in FlexInterval!S[] intervals)
+auto flex(S, F0, F1, F2)
+               (in F0 f0, in F1 f1, in F2 f2,
+                S[] cs, S[] points, S rho = 1.1)
+    if (isFloatingPoint!S)
 {
     import std.math: exp;
```
9il, line 140: use internal math

Issue #258, opened 2016-07-17
SDL is buggy for DUB


9il:
... so we need to change it to JSON. For example, `"dflags-ldc": ["-mcpu=native"]` SDL analog seems broken

wilzbach:
> For example, "dflags-ldc": ["-mcpu=native"] SDL analog

What did you try?

It should be something like `dflags-ldc: "-mcpu=native"`

But are you sure `dflags-ldc` exists? A [search](https://github.com/dlang/dub/search?utf8=%E2%9C%93&q=dflags-ldc) at the Dub repo didn't find it.

9il:
it exists for json, but not for sdl

9il:
`dflags-ldc "-mcpu=native"` does not work

s-ludwig:
Just stumbled over this - the eqiuvalent is `dflags "-mcpu=native" platform="ldc"`. See also http://code.dlang.org/package-format?lang=sdl#platform-specs

9il:
@s-ludwig Thank you, my bad

PR #279, opened 2016-08-04
Split up plotting examples into single files

wilzbach:
@9il I think I found a good way to restructure the plotting code, s.t. it's more user-friendly
## Compile & run a single plot

I put every plotting file into it's its own file that's runnable & executable. This should make it a lot easier to get an overview as one doesn't need to change a 700 LoC file. Of course all options can be modified directly by the user in a single example plot file.
They all use the new single file dub syntax and can be run like this:

```
./examples/flex_plot/test_normal.d
```
## Compile & run all plots

Also supports the dub single file syntax:

```
./examples/flex_plot.d
```

The common plotter also provides convenience CLI options that can be used to manipulate some plotting options on the CLI manually or only select a subset of plots.
If wanted I can move the CLI parsing also to common (it could be useful, but the single files are pretty small)

There are still some minor things to be done, but I wanted to hear your opinion on this idea.


examples/flex_plot/test1.d
```diff
@@ -0,0 +1,46 @@
+#!/usr/bin/env dub
+/+ dub.sdl:
```
9il, line 2: Should be json, see #258

wilzbach:
@9il seems like PyD doesn't support Python3.5 - see https://github.com/ariovistus/pyd/issues/50 for details :/
My workaround is to patch the `dub.json` from `appveyor.yml` - do you have a better idea what we could do for Windows users? (I just hope that nobody in their sane mind tries to use Windows)

wilzbach:
@9il this is finally passing on CircleCi and AppVeyor :tada: 
Can I go ahead, merge and improve upon (e.g. the `sdl` to `json` thing is a really minor thing)?

wilzbach:
Btw for the protocol, building Python 3.5 by source on CircleCi is needed as otherwise we get

```
/usr/bin/ld: cannot find -lpython3.5m
```

wilzbach:
> Btw for the protocol, building Python 3.5 by source on CircleCi is needed

Thanks to the caching at CircleCi it's adds only one minute extra to the build time.

PR #281, opened 2016-08-05
WIP Tinflex

wilzbach:
Contains the Tinflex code from #240, but squashed and rebased onto master.
Let's see whether CI passes ;-)
## TODO
- [x] replace stuff like z / 3 with z \* one_div_3
- [x] Do detailed paper work for cases +- 1 in Table 3. Need z^3 approximation instead of z^2. (better approximations will be sent as a separate PR once merged)
- [x] Fix bugs with in tinflexImpl
- [x]  clean code, add docs
- [x] remove commented tests
- [x] fix code coverage
## Once merged
- [x] benchmark against existing algorithms (dstats, cpp) in performance and quality
- [ ] publish second part of the blog post ([part 1](http://blog.mir.dlang.io/random/2016/08/19/intro-to-random-sampling.html), [part 2](https://github.com/libmir/blog/pull/5))
- [ ] send approximation PRs (for `area` and `flexImpl`)
- [ ] figure out how to deal with infinity and `c > 0`
## Known Issues
- For `infinity` intervals if `c > 0`, the interval type can't be determined anymore -> `undefined`
- We could recycle the uniform random variable generated with the discrete distribution sampler.
## Open question @9il 
- expose `pdf` to the user -> directly construct `Flex` object in the example plotting code
- how should we handle the specializations that are never used/executed (see comments below)?
- I added the CI testing of our plotting code on all three CIs (e.g. install python3.5. + matplotlib). While this was quite a work, I am not so sure anymore whether we should do this (increase in the build time ~1min, CircleCi ~1.5min, Travis ~ 1min, but ~4min for macOS (the font-cache for matplotlib is regenerated every time and I haven't figured out how to save/cache it)


source/mir/utility/linearfun.d
```diff
@@ -0,0 +1,117 @@
...
+    string toString() const
+    {
+        import std.format: format;
+        import std.math: abs, isNaN;
+        char sgn = intercept > 0 ? '+' : '-';
+        if (slope.isNaN)
+            return "#NaN#";
+        else
+            return format("%.2fx %c %.2f", slope, sgn, abs(intercept));
+    }
```
wilzbach, line 94: @9il this function is nice for debugging, however it has the downside that if one wants to copy the generated intervals this string is printed by default instead of e.g. `LinearFun!double(0.159263, -1.36003, 1.26786)`, so I am inclined to remove or rename this function. What's your opinion?

9il, line 94: ok

wilzbach, line 94: -> #282

wilzbach:
that's how it currently looks `n = 5K, rho =1.1`:

image
image
image
image

[All images at Google Drive](https://drive.google.com/open?id=0BwdiZp7qSaBheFJyQmJOU054V00)

examples/flex_plot/common/flex_common.d
```diff
@@ -0,0 +1,300 @@
...
+         S c, S[] points, S left = -3, S right = 3) const
+    {
+        auto cs = new S[points.length - 1];
+        foreach (ref d; cs)
+            d = c;
+        return plot(name, f0, f1, f2, cs, points, left, right);
+    }
+
+    /// ditto
+    auto plot(string name, in S function(S) f0, in S function(S) f1, in S function(S) f2,
```
wilzbach: I would prefer if I can pass in a `Tinflex` instance and don't  need to create these overloads.
This however requires that I am allowed to expose `pdf` for the `Flex` struct ;-)

source/mir/random/flex/package.d
```diff
@@ -0,0 +1,1012 @@
...
+    points = non-overlapping partitioning with at most one inflection point per interval
+    rho = efficiency of the Flex algorithm
+    apprMaxPoints = maximal number of splitting points before Flex is aborted
+
+Returns: Array of IntervalPoints
+*/
+FlexInterval!S[] flexIntervals(S, F0, F1, F2)
+                            (in F0 f0, in F1 f1, in F2 f2,
+                             in S[] cs, in S[] points, in S rho = 1.1,
+                             in int apprMaxPoints = 1_000, in int maxIterations = 1_000)
```
wilzbach: @9il imho we should expose either `apprMaxPoints` or `maxIterations` - which one do you prefer?

source/mir/random/flex/package.d
```diff
@@ -0,0 +1,1012 @@
...
+
+        logf("squeezeArea %(%f, %)", ipsD.map!`a.squeezeArea`);
+        version(Flex_logging_hex)
+            logf("squeezeArea %(%a, %)", ipsD.map!`a.squeezeArea`);
+
+        log("----");
+    }
+
+    // Flex is not guaranteed to converge
+    for (auto i = 0; i < maxIterations && nrIntervals < apprMaxPoints; i++)
```
wilzbach: If the code is correct `maxIterations` = `approxMaxPoints - startingPoints.length`. I think it's more intuitive for the user to specify only `approxMaxPoints`. @9il if you don't object I will remove `maxIterations`.

source/mir/random/flex/internal/area.d
```diff
@@ -0,0 +1,1024 @@
...
+                area /= sh.slope;
+            }
+        }
+    }
+
+    // if we receive an invalid value, we require the interval to be split
+    if (!isFinite(area))
+        area = S.max;
+    else if (area < 0)
+        area = S.max;
```
wilzbach, line 424: this safety check is never executed, so while being possible in theory I lack an example where it actually happens

source/mir/random/flex/internal/calc.d
```diff
@@ -0,0 +1,76 @@
...
+    {
+        if (rx < -S(1e3) || lx > S(1e3))
+            return 2 / (1 / lx + 1 / rx);
+
+        immutable d = atan(lx);
+        immutable b = atan(rx);
+
+        assert(d <= b);
+        if (b - d < S(1e-6))
+            return S(0.5) * lx + S(0.5) * rx;
```
wilzbach: TODO: these specializations are never tested/executed

source/mir/random/flex/internal/types.d
```diff
@@ -0,0 +1,346 @@
...
+        }
+        else if (iv.lt1x <= R && R <= iv.rt1x)
+        {
+            if (iv.lt2x < 0 && iv.rt2x > 0)
+                return T3a;
+            if (iv.lt2x > 0 && iv.rt2x < 0)
+                return T3b;
+        }
+
+        return undefined;
```
wilzbach: TODO @wilzbach: create a scenario where this can happen

source/mir/random/flex/internal/types.d
```diff
@@ -0,0 +1,346 @@
...
+
+        // TODO: zero seems to be a special case here
+        //assert(dt(-PI, 0) == T4a); // should be convex!
+
+        // but:
+        //assert(dt(PI, 2 * PI) == T3b);
+
+        //assert(dt(0, PI) == T4b); // should be concave (a)
+        // but:
+        //assert(dt(2 * PI, 3 * PI) == T2a);
```
wilzbach: TODO @wilzbach: clean the tests block for `determineType`

source/mir/random/flex/package.d
```diff
@@ -0,0 +1,1012 @@
...
+    // disallow NaN points
+    invariant {
+        import std.math : isNaN;
+        import std.meta : AliasSeq;
+        alias seq =  AliasSeq!(lx, rx, c, hatArea, squeezeArea);
+        foreach (i, v; seq)
+            assert(!v.isNaN, "variable " ~ seq[i].stringof ~ " isn't allowed to be NaN");
+
+        assert(lx < rx, "invalid interval - right side must be larger than the left side");
+    }
```
wilzbach: as there's no constructor the invariant isn't executed/tested anymore :/

wilzbach:
## Normal distribution

n = 10K, bins=50

image

n= 1M, bins= 500

image

with n =10M

image

9il:
Looks good)

source/mir/random/discrete.d
```diff
@@ -135,9 +137,10 @@ struct Discrete(T)
             arr[sm.index] = AltPair(sm.prob, lg.index);
 
             // update larger & reinsert
-            lg.prob += sm.prob - 1;
+            lg.prob += sm.prob;
+            lg.prob -= 1;
 
-            if (lg.prob < 1)
+            if (lg.prob < 1 || lg.prob.approxEqual(1, 1e-13, 1e-13))
```
wilzbach: This was added to increase the stability of the Discrete sampling when seeded with a static seed, but using different FP maths (e.g. `double` vs. `real`). I think it's a fair assumption to say that everything that's closely 0 is zero?

9il: No need. This is ad-hoc for Tinflex testing and it should not be a part of discrete

wilzbach: > No need. This is ad-hoc for Tinflex testing and it should not be a part of discrete

Well I just thought it's nice that if one writes tests with a seeded test for his algorithm the "randomness" should go away, e.g. sth like the following could yield _different_ results depending on 32/64.

```
auto ds = Discrete!double(0.25, 0.5, 0.25);
auto gen = Mt19937(42);
ds(gen);
```

Should we maybe add it as a template argument?

9il: no

9il: we will have fp pragmas

wilzbach: > we will have fp pragmas

... but what should I do for now? Just enable the sampling test for `x86_64` and `double`?

wilzbach:
@joseph-wakeling-sociomantic @WebDrake @9il 

I would like to share the draft version of my report at:

http://files.wilzbach.me/mir/reports/tinflex.pdf

First of all: the report is not final yet - it still contains many typos / grammars and misses structure at some pieces (-> consider it as "early draft"), but i wanted to share it with s.t. we are all on the "same page".

The source code for it (with all the scripts to generate the plots & the example D scripts) is also on [Github](https://github.com/wilzbach/flex-paper).
- what is the goal? 
  - chapter 1: explain the underlying ideas of the algorithm to the readers and motivate them to play with it (in fact I plan to use this as basis for the blog article that I have to submit to Google & to the D community as it hopefully will explain the problem visually.)
  - chapter 2-4: explain the exact code that is used in the algorithm (more technical)
- the scripts in the first chapter are [runnable](https://github.com/wilzbach/flex-paper/tree/master/samplers) (in fact same figures have been created with these plots) and automatically create histograms and cumulative histograms via [matplotd](https://github.com/wilzbach/matplotd) - a simple abstraction over PyD and matplotlib. Currently it's pretty dumb and only supports histograms, but I still think that's a very convenient way to plot. @9il in case you have better plans for this, I am happy to transfer it to libmir and make a proper structure
- while writing this document I also started to add more comments to the critical methods (they are currently in the benchmark PR, but I will merge them here soon)
- for chapter 3 I am rather sure that the equations are correct as I temporarily [added](https://github.com/wilzbach/mir/blob/23d68fd024c0be95df34d5ed8b20b058299f5443/source/mir/random/flex/internal/area.d#L359) them with `approxEqual` to our setup.

=> Let me know if you find anything that's
- missing (e.g. what is not explained there, what you expected to see before you opened the doc)
- unclear/misleading, ...
- other ideas

In this draft stage there's absolutely no need to go through the equations in detail, but @9il I think soon we should do this as I think I found at least four issues
1) some equations in `area.d` can be simplified (= less to compute)
2) There is one approximation in `area.d` [that looks suspicious](https://github.com/wilzbach/mir/blob/23d68fd024c0be95df34d5ed8b20b058299f5443/source/mir/random/flex/internal/area.d#L342)
3) Most approximations are done to avoid the fact that `b` (aka `slope`) can be `0`. Imho a quite reasonable approximation is to use sth. like `b = 0.0001` instead as we don't need to approximate then and thus keep a higher precision
4) The approximations in `flexImpl` don't match with my results (which are the same as we use these for the general case where `b > 1e6`, but I have to dig deeper on this one)

9il:
@wilzbach I would need few days for review

.travis.yml
```diff
@@ -24,6 +30,43 @@ script:
...
+ - ls ~/.matplotlib
+ - ls -al ~
+ # check that we created a valid file
+ - if [ ${TRAVIS_OS_NAME} = "osx" ] ; then
+     if [ $(stat -f%z plots/dist_normal_hs.pdf) -le 20000 ] ; then exit 1; fi ;
+   else
+     if [ $(stat -c%s plots/dist_normal_hs.pdf) -le 20000 ] ; then exit 1; fi
+   fi
+ # this file always changes (forces to repack the cache)
+ - rm -f $HOME/miniconda/lib/python3.5/__pycache__/_sysconfigdata.cpython-35.pyc
```
wilzbach: @9il Checking in the CI whether the plotting works is a bit complicated. and I think I should remove this from the mir setup. I just slows down the CI builds and such testing can be done in the upcoming matplotd:

https://github.com/wilzbach/matplotd

dub.json
```diff
@@ -16,6 +16,10 @@
         {
             "name": "dynamic",
             "targetType": "dynamicLibrary",
+        },
+        {
+            "name": "unittest",
+            "versions": ["to_be_filled"],
```
wilzbach, line 22: this is added, s.t. the version string can be conveniently changed, e.g. by the CI or user
(I would prefer if  DUB to allowed this via the CLI though)

examples/flex_plot/flex_common/flex_common/hist.d
```diff
@@ -0,0 +1,132 @@
+module flex_common.hist;
+
```
wilzbach, line 2: This developed itself into https://github.com/wilzbach/matplotd/blob/master/source/matplotd/hist.d and once matplotd is "a bit" stable I will change the code to use the library instead ;-)

wilzbach, line 2: edit: (this is a small, not so important side-project, so it won't happen soon)

source/mir/random/discrete.d
wilzbach: > we will have fp pragmas
> ... but what should I do for now? Just enable the sampling test for x86_64 and double?

I think it will take 2-6 months until we get the pragma, so what should I do for now?

wilzbach:
preview docs are up at http://preview.mir.rocks/mir_random_flex.html

WebDrake:
> preview docs are up at http://preview.mir.rocks/mir_random_flex.html

Nice job!  I would suggest putting a reference to the paper that introduces Tinflex early on in those docs (I would suggest at the end of your 1st paragraph).

.travis.yml
9il: Ok

wilzbach:
> Nice job! I would suggest putting a reference to the paper that introduces Tinflex early on in those docs (I would suggest at the end of your 1st paragraph).

Thanks - actually I wanted to link to the two posts on the new Mir Blog as they try to explain the topic for normal humans, but publishing the second post depends on this PR being merged, so I haven't inserted them yet.

http://blog.mir.dlang.io/random/2016/08/19/intro-to-random-sampling.html
https://github.com/libmir/blog/pull/5

edit: In any case I agree that the docs aren't perfect yet, but this PR is growing & growing, so I would argue to improve them in a follow-up PR.

PR #286, opened 2016-08-09
Tinflex - cumulative plotting

wilzbach:
@9il this adds cumulative histogram (with reference density) and the reference density to the normal histograms. Apart from that this PR includes more cleanups and a small bug fix.
It should be noted that the histogram scaling for the non-cumulative case isn't perfect - I just scale the pdf according to the value of the largest bins.

In any case I think the cumulative plots make it easier to see if something is going wrong, here are a couple:

image

image

image

image

image

I dumped [all plots](https://drive.google.com/open?id=0BwdiZp7qSaBhNlUxeVZSTXl3cHM) to Google Drive.


wilzbach:
Btw about the last plot I think that was just a problem of having to less bars:

n=500K, bars=200

image

n=5M, bars=500

image

PR #287, opened 2016-08-10
Flex2 - first benchmark

wilzbach:
Hi @9il @joseph-wakeling-sociomantic @WebDrake

As the (Tin)flex algorithm is about to go into its final stage (@WebDrake see e.g. #286 and #281), I did a quick performance benchmark across existing tools. Note that this is benchmark assumes that all samplers return distributions with "good random quality", but assessing that will be another benchmark.

1) `flexNormal` isn't the fastest as it needs at least three random variables (in their C code the author manage to [reuse one](https://github.com/cran/Tinflex/blob/master/src/Tinflex.c#L218) of them, but I am currently not a huge fan of leaking such details from the new Alias method sampler), however the point of the (tin)flex algorithm isn't to be the fastest sampler, but to allow easy construction of new distributions. That being said we should of course try to squeeze the last wasted cycles out of it.
2) I hope it shows how valuable Ziggurat is/will be (>3x faster & there's still space for further optimizations)

About the benchmark itself:
- as the rest of mir it requires the latest `ldc` (> 1.0.0)
- the values below sampled 10M values, but the default (1M) works fine to show the differences as well
- @WebDrake I don't know why `hap` is so slow, but I guess it's due to `uniform01`
- the three different versions of the flex algorithm use different `rho` and thus higher/lower efficiency (-> number of times it needs to resample for one value)

```
dub run --build=release-nobounds --compiler=ldmd2-git --single benchmarks/flex/normal_dist.d 
```

```
boxMueller.naive   = 1 sec, 379 ms, 411 μs, and 3 hnsecs
boxMueller.hap     = 23 secs, 568 ms, 35 μs, and 4 hnsecs
boxMueller.dstats  = 1 sec, 217 ms, 655 μs, and 3 hnsecs
boxMueller.atmos   = 2 secs, 113 ms, 476 μs, and 1 hnsec
flexNormal.slow    = 2 secs, 460 ms, 365 μs, and 9 hnsecs
flexNormal.medium  = 2 secs, 151 ms, 805 μs, and 3 hnsecs
flexNormal.fast    = 2 secs, 27 ms, 822 μs, and 1 hnsec
ziggurat           = 423 ms, 150 μs, and 9 hnsecs
```

Just in case here are the links to the other packages
- [dstats](https://github.com/DlangScience/dstats/blob/master/source/dstats/distrib.d)
- [hap](https://github.com/WebDrake/hap/blob/master/source/hap/random/distribution.d)
- [atmosphere](https://github.com/9il/atmosphere/blob/master/source/atmosphere/random.d)
- [Ziggurat](https://github.com/libmir/mir/pull/261)

### DIFF
--- a/benchmarks/flex/normal_dist.d
+++ b/benchmarks/flex/normal_dist.d
@@ -22,13 +22,14 @@ LDC - the LLVM D compiler (798cda):
 
 $ dub run --build=release-nobounds --compiler=ldmd2-git --single benchmarks/flex/normal_dist.d
 
-boxMueller.naive   = 12 secs, 36 ms, and 998 μs
-boxMueller.hap     = 40 secs, 844 ms, 230 μs, and 6 hnsecs
-boxMueller.dstats  = 8 secs, 4 ms, 556 μs, and 4 hnsecs
-flexNormal.slow    = 14 secs, 54 ms, 592 μs, and 4 hnsecs
-flexNormal.medium  = 12 secs, 13 ms, 252 μs, and 9 hnsecs
-flexNormal.fast    = 11 secs, 174 ms, 573 μs, and 5 hnsecs
-ziggurat           = 3 secs, 538 ms, 75 μs, and 9 hnsecs
+boxMueller.naive   =   850 ms, stddev: 82  ms
+boxMueller.hap     =  3618 ms, stddev: 4   ms
+boxMueller.dstats  =   801 ms, stddev: 2   ms
+boxMueller.atmos   =   934 ms, stddev: 2   ms
+flexNormal.slow    =  1406 ms, stddev: 4   ms
+flexNormal.medium  =  1206 ms, stddev: 1   ms
+flexNormal.fast    =  1124 ms, stddev: 0   ms
+ziggurat           =   357 ms, stddev: 5   ms
 +/
 
 import mir.random.flex;
@@ -97,7 +98,8 @@ void main()
     auto flexNormalMedium = genNormal!S(1.1);
     auto flexNormalFast = genNormal!S(1.0001);
 
-    auto zigguratNormal = normal!(S, uint)();
+    // mir ziggurat is temporarily excluded as it hasn't been merged yet
+    //auto zigguratNormal = normal!(S, uint)();
 
     // just pick any rng gen, it will have the same speed for all algorithms
     import std.random : Mt19937;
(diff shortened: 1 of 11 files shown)


9il:
looks good

benchmarks/flex/normal_dist.d
```diff
@@ -0,0 +1,326 @@
...
+    // just pick any rng gen, it will have the same speed for all algorithms
+    import std.random : Mt19937;
+    auto gen = Mt19937(42);
+
+
+    auto bench = benchmark!(
+        { r += boxMueller!S(0, 1, gen); },
+        {
+            import hap.random.distribution : normal;
+            r += normal(0, 1, gen);
```
WebDrake: From `hap.random` docs:

> Note that this function uses two variates from the uniform random number generator to generate a single normally-distributed variate. It is therefore an inefficient means of generating a large number of normally-distributed variates. If you wish to draw many variates from the normal distribution, it is better to use the range-based `normalDistribution` instead.

For fair speed comparison, you need to be creating a `NormalDistribution` range.  That should only be initialized once, i.e.

``` D
auto hapNorm = normalDistribution(0, 1, gen);
```

... & then in your benchmark, use `hapNorm.front` and `hapNorm.popFront()` (you don't pass in the RNG on a per-call basis).

WebDrake: It's also possible that your call to `normal` here is under the hood initializing a new `NormalDistribution` instance per call, which could explain some of the additional slowdown.

wilzbach:
@WebDrake

> From hap.random docs:

Sorry :/

> For fair speed comparison, you need to be creating a NormalDistribution range. That should only be initialized once, i.e.

To have a "fair" benchmark I also changed the code to use multiple runs and only show the average between all runs (& for comparison the stddev).

parameters are are `20` runs and 10M samples.

```
boxMueller.naive   =   850 ms, stddev: 82  ms
boxMueller.hap     =  3618 ms, stddev: 4   ms
boxMueller.dstats  =   801 ms, stddev: 2   ms
boxMueller.atmos   =   934 ms, stddev: 2   ms
flexNormal.slow    =  1406 ms, stddev: 4   ms
flexNormal.medium  =  1206 ms, stddev: 1   ms
flexNormal.fast    =  1124 ms, stddev: 0   ms
ziggurat           =   357 ms, stddev: 5   ms
```

wilzbach:
FYI I temporarily removed ziggurat from the benchmark until it finds its way into Mir ;-)

PR #301, opened 2016-08-22
[Tinflex] Always use left secant

wilzbach:
While the test are highly sensitive against changes atm, I think it shouldn't matter whether the left or right point is used to construct the interval. I open this here to keep track of it and make further tests (a first glance at the plots seems to approve my assumption).


9il:
It should be either left or right

9il:
Feel free to reopen if I am not wrong

PR #307, opened 2016-08-23
Add C++ to NormalDist benchmark

wilzbach:
I need to find a simple, convenient & fair way to allow running the benchmarks, hence WIP (for the blog I ran them manually)

@9il as you disliked my initial Makefile approach I will probably go for sth. similar like [kostaya/benchmark](https://github.com/kostya/benchmarks), I will post some ideas tonight.

### DIFF
--- a/benchmarks/flex/build.sh
+++ b/benchmarks/flex/build.sh
@@ -1,0 +1,5 @@
+#!/bin/bash
+
+file="normal_dist"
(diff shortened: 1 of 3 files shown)


9il:
I don't think we need this test. Different c++ libs may have different algorithms. In addition, the performance should be the same as D for the same algorithm

9il:
`mir-random` uses the same algorithm as CLANG++

PR #347, opened 2016-09-22
remove dmd support

9il:


wilzbach:
Hmm I always liked the fast compilation times of the DMD BE for testing :/

9il:
I too, however if a lot of code has `version(DMD)`,  then we need to test it with LDC anyway. Furthermore we can create an configuration to run LDC without code generation. 99% WIP errors are cached by FE.

BlackEdder:
> I too, however if a lot of code has version(DMD), then we need to test it with LDC anyway. Furthermore we can create an configuration to run LDC without code generation. 99% WIP errors are cached by FE.

I think it is a real shame not at least to support compilation with DMD. And I don't understand the reasoning above. It is trivial to make travis run both LDC and DMD test builds.

WalterBright:
I'd prefer to see the DMD support put back in and we can work to remove the conditional compilation sections.

9il:
It will slowdown Mir development because a number of conditions will grow fast and I would need to test new workarounds for DMD.

`v0.17.0-alpha9` can be used with DMD, it has all conditions, which was removed in this PR.

WalterBright:
I suggest instead of workarounds we can add to dmd what is missing so the same code compiles for both. All I need from you is Enhancement Requests for those items.

(Also, workarounds can be abstracted away in the meantime, especially if it is the same set of workarounds over and over.)

9il:
> I suggest instead of workarounds we can add to dmd what is missing so the same code compiles for both. All I need from you is Enhancement Requests for those items.

Thank you! Will do

9il:
> All I need from you is Enhancement Requests for those items.

This one looks like hardest to implement:
https://issues.dlang.org/show_bug.cgi?id=16489

PR #357, opened 2016-09-24
enable -singleobj

MartinNowak:
- improves cross module inlining

### DIFF
--- a/benchmarks/glas/gemm_report.d
+++ b/benchmarks/glas/gemm_report.d
@@ -3,9 +3,9 @@
 {
 	"name": "gemm_report",
 	"dependencies": {"mir": {"path": "../.."}, "cblas": "~>0.2.0"},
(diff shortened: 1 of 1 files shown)


MartinNowak:
Mmh, dub passes this as well to ldmd, though it's dflags-ldc.
And ldc complains about the duplicate flag.

thewilsonator:
I got a duplicate flag when trying to use -oq flag, but ldc didn't complain. A dub bug perhaps?

dnadlinger:
LDC should probably allow multiple mentions of `-singleobj`, in line with how most other DMD-style flags are handled. LLVM disallows duplicate flags by default, and we haven't ever made a point out of ensuring they are allowed everywhere (or the decision to do so, for that matter).

When building executables through LDMD, it is already the default, though, and will also be for straight `ldc2` starting from the next release.

